PR #411, opened 2026-06-16
docs(rav1d-safe): state frame_size_limit unit/default + cancellation + types

lilith:
## Why

An "insulated external developer" usability test — given **only** rav1d-safe's README and asked to build an untrusted-AV1 decode server — surfaced four gaps. The most important: the README's security-guard story for untrusted input (`frame_size_limit`) never stated its **unit** or **default**, so a developer couldn't size the DoS guard correctly. All claims below were verified against `src/managed.rs` and `src/obu.rs`.

## What changed (README + one CHANGELOG line, no code)

1. **`frame_size_limit` — unit + default now stated precisely.** It is the max **`width * height` in pixels** (not bytes, not max-dimension; verified at `src/obu.rs:2344` — `width * height > frame_size_limit`), and the **default is `120_000_000` (120 MP)** — *not* unlimited and *not* the "4K" the old example comment implied (`src/managed.rs:187`). Documented `0` = disabled, the 32-bit `8192*8192` clamp (`src/lib.rs:161`), and that rejection surfaces as `Error::Other` carrying `ERANGE` (the `From<Rav1dError>` `_` arm — **not** `InvalidData`).

2. **Cancellation reality.** Verified there is **no** stop/cancel/abort/`enough` token anywhere in `src/` — added an honest **Cancellation** section: no in-flight decode abort exists; `frame_size_limit` is the only pre-decode guard (and it bounds declared dimensions, not wall-clock time). Servers must wrap the decoder with their own timeout. Flagged as a tracked gap.

3. **Types / canonical import + reconciled `Planes` naming.** Added one canonical `use rav1d_safe::{…}` listing the full crate-root surface (verified against the `pub use` in `lib.rs:202`). Reconciled the table-vs-example mismatch: `Planes::Depth8`/`Depth16` are **enum variants**; `Planes8`/`Planes16` are the **structs** they wrap — both are real and re-exported.

4. **Output format.** Documented that a decoded `Frame` is **planar YUV** (no built-in RGB), with the `Planes8/16` → `y()` / `u()` / `v()` (Option, `None` for I400) accessors, `PixelLayout` subsampling (`I400/I420/I422/I444`), and 8 vs 10/12-bit (`Depth8` = u8, `Depth16` = u16). Accessor names verified at `src/managed.rs:729/752/779`.

The existing (excellent) Input-Format section is untouched.

## Note for maintainers

`src/managed.rs:136`'s inline doc claims frame-size rejection yields `Err(InvalidData)`, but the `From<Rav1dError>` impl maps `ERANGE` via the catch-all to `Error::Other` — so the README documents the *actual* `Error::Other` behavior. The inline source doc comment is left as-is (out of scope for a README-only change) but flagged here.

Found via an insulated external-developer usability test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ All notable changes to the `rav1d-safe` crate are documented in this file. Forma
   within overlapping spreads); full md5 conformance suite green.
 
 ### Changed
+- **README accuracy pass for untrusted-input servers** (found via an insulated external-developer usability test): documented that `frame_size_limit` is a **total-pixel** (`width * height`) cap with a **120 MP default** (not unlimited, not the "4K" the old example implied) and that `0` disables it; added an honest **Cancellation** note that no in-flight decode-abort token exists (the frame-size cap is the only pre-decode guard); added a canonical crate-root `use rav1d_safe::{…}` import and reconciled the `Planes` variant names (`Planes::Depth8`/`Depth16` enum variants vs `Planes8`/`Planes16` structs); and documented the **planar-YUV output** format (`y()`/`u()`/`v()` accessors, `PixelLayout` subsampling, 8 vs 10/12-bit planes) (#411).
 - **Test runner switched to `cargo-nextest`** (justfile + CI). Nextest runs each
   test in its own process, which removes a whole class of cross-test global-state
   races structurally — the archmage token-permutation vs. v4x parity race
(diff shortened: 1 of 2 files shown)


lilith:
Filed the cancellation capability gap (gap #2) as #412 — this PR documents the reality (no in-flight decode abort; `frame_size_limit` is the only pre-decode guard); #412 tracks adding a cooperative-cancellation token.

PR #416, opened 2026-06-25
Implement DisjointMut creation from a mutable slice

Shnatsel:
Turns out this is possible under both Stacked Borrows and Tree Borrows. This _feels_ like crimes against provenance, but `miri` says this is totally legit. 

Under Stacked Borrows, “a reference exists somewhere” is not by itself UB. The problem would be accessing/reborrowing the full slice through that `&mut [V]` while `&mut` guards to subranges exist. That would retag the whole slice and invalidate the chunk borrows. So we carefully avoid materializing `&[V]`/`&mut [V]` for the full slice by using raw pointers in the creation instead.

Provenance of the included unit test looks like this:

1. `DisjointMut::new(&mut buf[..])` consumes the unique full-slice borrow.
2. The stored `&mut [V]` acts as the provenance source.
3. `index_mut(0..32)` creates a child `&mut` for only `0..32`.
4. `index_mut(32..64)` creates another child `&mut` for only `32..64`.
5. The original full-slice `&mut` remains dormant until those children are gone.

This is also (roughly) how [`slice::split_at_mut()`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.split_at_mut) in the standard library deals with provenance, but we have to go to slightly greater lengths to avoid retagging.

### DIFF
--- a/crates/rav1d-disjoint-mut/src/lib.rs
+++ b/crates/rav1d-disjoint-mut/src/lib.rs
@@ -323,6 +323,7 @@ mod sealed {
     pub trait Sealed {}
 
     impl<V: Copy> Sealed for Vec<V> {}
+    impl<V: Copy> Sealed for &mut [V] {}
     impl<V: Copy, const N: usize> Sealed for [V; N] {}
     impl<V: Copy> Sealed for [V] {}
     impl<V: Copy> Sealed for Box<[V]> {}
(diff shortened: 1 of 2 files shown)


crates/rav1d-disjoint-mut/src/lib.rs
```diff
@@ -1542,6 +1543,32 @@ unsafe impl<V: Copy> AsMutPtr for Vec<V> {
...
 }
 
+/// SAFETY: Copies the stored mutable slice reference as a raw slice pointer
+/// without materializing a reference to the full slice. The data is borrowed
+/// from the caller, so creating `&[V]` or `&mut [V]` for metadata lookup would
+/// conflict with outstanding guards under Stacked Borrows.
+unsafe impl<V: Copy> AsMutPtr for &mut [V] {
+    type Target = V;
+
+    unsafe fn as_mut_slice(ptr: *mut Self) -> *mut [Self::Target] {
```
danielhenrymantilla, line 1553:
```suggestion
    unsafe fn as_mut_slice(ptr: *mut Self) -> *mut [V] {
```

to be consistent with the signature of `as_mut_ptr()` (in this direction since it seems more readable, which is quite important when around `unsafe` code)

danielhenrymantilla (approved):
LGTM ✅, soundness-wise.

  - Usage of `PtrOrRefTo<&mut _>` does indeed not reässert lack-of-aliasing of the `&mut`, and thus avoids invalidating outstanding pointers or whatnot derived from it.

crates/rav1d-disjoint-mut/src/lib.rs
```diff
@@ -1542,6 +1543,32 @@ unsafe impl<V: Copy> AsMutPtr for Vec<V> {
...
+        unsafe { ptr.cast::<*mut [V]>().read() }
+    }
+
+    unsafe fn as_mut_ptr(ptr: *mut Self) -> *mut V {
+        // SAFETY: Same as `as_mut_slice`; casting the raw slice pointer to its
+        // data pointer does not create an intermediate reference.
+        unsafe { Self::as_mut_slice(ptr).cast() }
+    }
+
+    fn len(&self) -> usize {
```
danielhenrymantilla, line 1565: If we wanted to be very pedantic (and around this subtle code, we may!), we could be spelling out `Self` in these signatures each time:


```suggestion
unsafe impl<'r, V: Copy> AsMutPtr for &'r mut [V] {
    type Target = V;

    unsafe fn as_mut_slice(ptr: *mut &'r mut [V]) -> *mut [V] {
        // SAFETY: `&mut [V]` and `*mut [V]` have the same pointer layout.
        // Reading the raw pointer value avoids reborrowing the full slice.
        unsafe { ptr.cast::<*mut [V]>().read() }
    }

    unsafe fn as_mut_ptr(ptr: *mut &'r mut [V]) -> *mut V {
        // SAFETY: Same as `as_mut_slice`; casting the raw slice pointer to its
        // data pointer does not create an intermediate reference.
        unsafe { Self::as_mut_slice(ptr).cast() }
    }

    fn len(self: &&'r mut [V]) -> usize {
```

lilith:
The downstream use cases on rav1d didn't materialize since owned vecs turns out to be needed with the multithread architecture

Shnatsel:
I didn't end up using this in my project either because it does a lot of indexing and so I end up in the slow spinning path a lot, which tanks performance. Collecting the ranges into a `Vec<&mut [T]>` in advance turned out to be faster than this kind of tracking due to the high contention.

lilith:
Yeah I did the same for zenjxl-decoder, it worked well

PR #445, opened 2026-08-06
fix(tile-threading): stop reserving inter-row gaps in the strided block guard

lilith:
## The bug

Single-frame multithreaded decode panics on real content. `mt_stress_4k` fails at `threads>=2` inside one trial:

```
overlapping DisjointMut:
 current: &mut _[0..57616] at src/safe_simd/itx_arm.rs:8589
existing: &mut _[24000..24016] at src/ipred.rs:515
```

## Root cause

`narrow_guard_mut` (behind `strided_slice_mut`) reserves **one contiguous range** spanning `(h-1)*stride + w` pixels — the whole extent from a block's first pixel to its last, **including the inter-row gaps**, which belong to other blocks. On a 3840-wide frame a 16×16 block reserves 57,616 bytes in order to write 256.

That is fine single-threaded and wrong under tile threading, where AV1 tiles partition a frame **by columns**: two workers routinely write the same rows at different columns. The second lands inside the first's reserved gap and the tracker panics. The two writes are genuinely disjoint — a false positive. In `unchecked` builds the same over-broad reservation just isn't checked.

This is the zenavif#30 class, and that fix was only ever applied to the loopfilter and `ipred`, which grew their own `compact_*_per_row` paths. Everything else still reserves gaps: **itx_arm 13, mc_arm 15, cdef 3, itx 2, mc 1, ipred_arm 1**.

## The fix

`WithOffset::block_mut` generalises the compact approach behind one type instead of open-coding it a fourth time:

- **tile threading OFF** → exactly the previous behavior: one contiguous guard, the picture's stride, no copying, no added cost.
- **tile threading ON** → the block is copied into a compact `w×h` scratch buffer through **per-row guards**, each reserving only the `w` pixels actually touched, and written back per row on drop. No gap is ever reserved.

Callers take the stride from `BlockMut::byte_stride()` rather than the picture, since the compact buffer has its own.

Still `forbid(unsafe_code)` — no new `unsafe`.

## Scope

Converted itx_arm's 12 safe-build sites. The 13th is behind `#[cfg(feature = "asm")]`, hands a raw pointer to an asm kernel walking the picture stride, and is out of scope. **mc_arm / cdef / itx / mc / ipred_arm remain exposed** — mechanical conversions to the same helper, deliberately left to a follow-up rather than bundled into a fix that needs to be reviewable.

## Verified (aarch64 macOS, 12 cores)

- `mt_stress_4k` **passes at threads 1/2/4/8/16 × 5 trials**. Before: panics at threads=2, trial 0.
- **Decoded output is bit-identical.** Comparing every (vector, actual-md5) pair the suite emits, before vs after: **0 differ**.
- Pre-existing and unchanged: 5 `decode_cpu_levels` / `test_vq_suite_standalone` md5 failures reproduce with this change stashed, emitting the same `actual=` hashes. Not introduced here, not fixed here.

## Why this survived

`mt_stress_4k` reads `test-vectors/bench/photo_4k.avif` — a gitignored local asset generated by a justfile recipe needing an 8K source AVIF plus avifenc/avifdec/ImageMagick. The file was absent, so **the one test that catches this bug could not run.** Making that vector CI-provisionable is worth a follow-up.

Found while wiring in-flight decode cancellation in imazen/zenavif (PR #31), where cancelling a 4K decode requires multithreaded decode to work at all.

### DIFF
--- a/include/dav1d/picture.rs
+++ b/include/dav1d/picture.rs
@@ -84,11 +84,17 @@ use crate::include::common::bitdepth::BitDepth;
 
 /// Execute a closure with mutable byte access to a w×h pixel block.
 ///
-/// In single-threaded mode: zero-copy via `narrow_guard_mut`.
-/// In multi-threaded mode: copies to/from a compact buffer with per-row guards.
+/// The closure receives `(bytes, offset, stride)` — a mutable byte slice, the
+/// BYTE offset to the first pixel, and the byte stride between rows.
 ///
-/// The closure receives `(bytes, offset, stride)` — a mutable byte slice,
-/// the byte offset to the first pixel, and the byte stride between rows.
+/// This is a closure-shaped front end for [`WithOffset::block_mut`], which owns
+/// the single-threaded-vs-tile-threaded policy. The two exist because call sites
+/// come in two shapes — a closure suits the x86-64 and generic dispatchers, an
+/// RAII guard suits the aarch64/wasm ones that need the slice to outlive a
+/// borrow — but there must be exactly ONE implementation of the policy behind
+/// them. When there were two, the aarch64 half silently kept reserving inter-row
+/// gaps long after the x86-64 half stopped, which is the bug this whole change
+/// exists to fix.
 #[inline]
 pub fn with_pixel_guard_mut<BD: BitDepth, R>(
     pic: &crate::src::with_offset::WithOffset<&Rav1dPictureDataComponent>,
@@ -96,21 +102,17 @@ pub fn with_pixel_guard_mut<BD: BitDepth, R>(
     h: usize,
     f: impl FnOnce(&mut [u8], usize, isize) -> R,
 ) -> R {
-    use crate::src::strided::Strided as _;
-    let pixel_size = core::mem::size_of::<BD::Pixel>();
-    if tile_threading_active() {
-        let (mut buf, byte_stride) = pic.compact_read_per_row::<BD>(w, h);
-        let result = f(&mut buf, 0, byte_stride as isize);
-        pic.compact_write_back_per_row::<BD>(w, h, &buf);
-        recycle_compact_scratch(buf);
-        result
-    } else {
-        let (mut guard, base) = pic.narrow_guard_mut::<BD>(w, h);
-        let bytes = guard.as_mut_bytes();
-        let offset = base * pixel_size;
-        let stride = pic.stride();
-        f(bytes, offset, stride)
-    }
+    let mut block = pic.block_mut::<BD>(w, h);
+    // `BlockMut::base()` is a PIXEL index — 0 for a compact block, and the
+    // last row's index for a direct block on a negative-stride picture. This
+    // closure's contract is a BYTE offset, so the scaling happens here rather
+    // than inside `BlockMut`, whose own users index the byte slice directly.
+    let offset = block.base() * core::mem::size_of::<BD::Pixel>();
+    let stride = block.byte_stride();
+    // `block` is dropped at the end of this expression, i.e. after `f` has run:
+    // a compact block therefore writes back before the caller observes `R`,
+    // exactly as the previous open-coded version did.
+    f(block.as_mut_bytes(), offset, stride)
 }
 
 /// Execute a closure with read-only byte access to a w×h pixel block.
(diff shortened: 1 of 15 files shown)


lilith:
### Note on the Clippy / Format CI failures

Both were **already failing on `main`** — they are not introduced by this PR. The `CI` workflow (which owns the Clippy and Format jobs) only triggers on pull requests, so pushes to `main` never gate on it and the debt accumulated unnoticed.

**Format** — one over-long line in `src/cdef.rs`, a file this PR does not otherwise touch. Fixed here as a separate `style: cargo fmt` commit so it stays reviewable apart from the fix.

**Clippy** — still red, and deliberately left alone. Reproduced with `cargo clippy --no-default-features --features bitdepth_8,bitdepth_16 -- -D warnings` (the exact CI invocation). Every finding is pre-existing:

- `COS_PI_1_16` … `COS_PI_7_16`, `DCT4_C1` … `DCT4_C4` — each occurs exactly **once** in `itx_arm.rs`, i.e. at its own definition. Dead before this branch.
- `inv_txfm_add_dct_dct_{4x4,8x8,16x16}_{8,16}bpc_inner`, `iclip_diff`, `signed_idx`, `loop_filter_core`, `lpf_{h,v}_sb_inner` — scalar reference implementations, unreferenced under this feature set.
- `clamp-like pattern without using clamp function` at `itx_arm.rs:7673` (`p.max(0).min(255)`) — inside a scalar reference function, untouched by this PR.

I did not fold these into this PR: several are "delete or `#[allow]`?" judgement calls about intentionally-kept reference code, which is the author's call, not a drive-by from a threading fix. Happy to do it as a follow-up if you want the lint debt cleared — and it may be worth running the `CI` workflow on `main` pushes so it cannot silently re-accumulate.

lilith:
### Scope update — the sweep is now complete for the safe build

A follow-up commit (`f42e968`) converts the remaining reachable sites, so this PR no longer leaves the defect half-fixed as the original description said.

**Converted (17 more):** `mc_arm.rs` 7 (the aarch64 `avg_`/`w_avg_`/`mask_`/`blend_`/`blend_dir_`/`w_mask_`/`mc_put_dispatch` wrappers), `mc_wasm.rs` 3, `itx_wasm.rs` 2, `cdef_wasm.rs` 2 (plus dropping the now-meaningless picture-stride parameter from `cdef_filter_block_simd_{8,16}bpc` and its 6 call sites).

**Remaining 9 sites are all behind non-default feature gates** (`asm` / `__simd_test`) — `cdef.rs` 3, `itx.rs` 2, `mc.rs` 1, `ipred_arm.rs` 1, `itx/part10_dispatch.rs` 1, `itx_arm.rs` 1. They hand a raw pointer plus the picture stride to an asm/FFI kernel that walks that stride, or (in the `__simd_test` A/B harness) deliberately save and restore the full strided span *including* gaps — a compact w×h view cannot express either, so converting them would be wrong, not merely unnecessary. Verified independently: `cargo check --no-default-features --features "bitdepth_8,bitdepth_16"` compiles, and **the safe-SIMD build now contains no gap-reserving guard site at all**.

**Supporting find:** every caller of the mc_arm wrappers (`avg_direct` … `mc_put_direct` in `src/mc.rs`) is `#[cfg(not(feature = "asm"))]`, so no compact buffer can reach an asm kernel. And the **x86_64 twins of those same `*_dispatch_inner` functions already route through `with_pixel_guard_mut`** — so mc_arm was simply the un-converted half of a pattern this repo had already established elsewhere. That is good evidence the approach here is the intended one rather than a new invention.

### Verification of the follow-up

- `mt_stress_4k` passes (4K decode at threads 1/2/4/8/16).
- **Bit identity: 2,100 `(tier, vector-id) -> actual-MD5` pairs compared before vs after — 0 differed.** Pass/fail counts and failing-test-name sets identical.
- `tile_threading_parity` (threads=1 vs threads=4 MD5 equality — the gate that would catch compact-path corruption specifically), `decode_md5_committed`, `safe_simd_crashes`: pass.
- Clippy diagnostic set and per-file counts identical before/after — no new warnings. `cargo fmt --check` clean. No `unsafe` added.

### Two honest caveats

1. **The three wasm files are compile-checked only** (`cargo check --target wasm32-unknown-unknown`, clean). They cannot execute on this box and the aarch64 suite does not build them. Their conversion is mechanically identical to the aarch64 ones that *are* runtime-verified, and the compact path is dormant on wasm anyway (no tile threading without threads) — but there is no runtime evidence for them.
2. Correcting a number in an earlier comment: the pre-existing failing-test count on this branch's baseline is **34**, not 5 — I had counted only one test binary. Unchanged before vs after, and all pre-existing.

PR #448, opened 2026-08-06
fix(cdef): aarch64 16bpc CDEF read the wrong pri_tap parity bit (#446)

### LINKED ISSUE #446: 12-bit inter-frame decode appears to diverge from libaom (inferred, not yet directly measured)
## Summary

A 12-bit AVIF animation decodes to different pixels through rav1d-safe than through a pure-Rust libaom port whose planes are **byte-identical to real libaom v3.14.1**. I have not yet compared rav1d-safe's planes to C directly, so this is an inference — but the chain is tight and the reproducer is exact.

**Vector:** `colors-animated-12bpc-keyframes-0-2-3.avif` (libavif animated set). Profile 2, 64x64, bd12, **4:2:2**, SB128, frames `KEY, INTER, KEY, KEY, INTER`.
**Diverging frame:** 1 — an inter frame. Frames 0/2/3 (keyframes) agree.

## The inference chain

1. zenavif decodes this file through two backends and byte-compares. Frame 1 differs; frame 0 of the same track agrees exactly.
2. Both backends feed the **same** conversion kernel with the same arguments — `zenavif::yuv_convert::yuv16_to_rgbx_strip` (`src/decoder_managed/aom.rs:423` and `src/decoder_managed/plane_convert.rs:620`). Frame 0 agreeing controls for the arguments being equivalent; a systematic argument mismatch would break frame 0 too.
3. The other backend (imazen/zenav1-aom) was measured against **real libaom v3.14.1** in-process (`aom_codec_av1_dx`): **0 of 100 cells diverge** — 40/40 animated frames, 48/48 `[KEY, P]` grid cells, byte-exact, identical under forced-scalar. Record: imazen/zenav1-aom#8 and its PR #9.

Therefore the difference lives in the **decoded planes**, and the other port's planes equal C's. That points here.

## Character of the difference

Measured on zenavif's final `Rgba&lt;u16&gt;` output (8 bytes/px), first 32768 bytes: 328 bytes differ. After `scale_pixels_to_u16` these work out to **±1/±2 code steps at 12 bits**, concentrated in the two chroma-driven channels (R←V, B←U); G mixes both with smaller coefficients and rounds to no change. So: a small chroma reconstruction difference on an inter frame at high bit depth — not a structural break.

## The settling check (what would confirm or refute this)

Hash rav1d-safe's decoded planes for this vector's frame 1 against the now-oracle-backed golden in zenav1-aom: `…12bpc…color.md5` line 2 = `9f2291bf3f75dd440bc1d64ae26e0ac8`. If they match, this issue is wrong and the divergence is elsewhere in zenavif; if they differ, it is here.

## Possibly related

This repo currently has **34 pre-existing failing tests** on `main` across `decode_cpu_levels` / `decode_md5_verify` / `decode_permutations` (stable, reproducible, same `actual=` hashes run to run). I have not checked whether any involve high bit depth or inter frames — if they do, this may be one symptom of a shared cause rather than an isolated bug. Worth checking before treating this as new.

## Reproducer

```
cd zenavif   # branch svtav1-rs-backend
just download-vectors
cargo test --features aom-backend,encode,encode-imazen \
  animations_decode_identically_across_backends
```

Found while wiring in-flight decode cancellation (PR #445). Filed here after I mis-filed it against zenav1-aom on a byte-delta histogram read — see imazen/zenav1-aom#8 for that correction.

lilith:
Closes #446. A real wrong-pixels defect, one line, aarch64-only, high-bit-depth-only.

## The bug

`src/safe_simd/cdef_arm.rs` selected the CDEF primary tap from **bit 0** of `pri_strength`:

```rust
let pri_tap = 4 - (pri_strength & 1);
```

But the caller has already scaled the level by `bitdepth_min_8` (`y_pri_lvl = (y_lvl >> 2) << bitdepth_min_8` in `cdef_apply`), so the tap-selecting parity bit lives at bit `bitdepth_min_8`, not bit 0. Every high-bit-depth block with an odd `pri_strength >> bitdepth_min_8` filtered with **pri_tap 4 instead of 3**.

**8-bit was immune** because `bitdepth_min_8` is 0 there — which is exactly why this survived: the whole 8-bit corpus agrees, and only high-bit-depth content diverges.

The scalar path (`src/cdef.rs`) and the x86 and wasm ports already did the shift. So does dav1d's own NEON path (`src/arm/64/cdef_tmpl.S`: `lsr w9, w3, w9` / `and w9, w9, #1` under `.if \bpc == 16`).

**`CpuLevel::Scalar` does not route around it** — aarch64 CDEF dispatch always returns `true` and consults no CPU mask, so all CDEF on ARM goes through this file.

## How it was established (not by elimination)

#446 was filed as an inference from two Rust decoders disagreeing. That is not evidence, so it was replaced with measurement:

1. Fed rav1d-safe the **identical `.obu` bytes** used by zenav1-aom's harness, removing container extraction as a variable, and hashed with that harness's exact recipe (one md5 per shown frame, Y then U,V, cropped dims, u16 LE).
2. Local **`aomdec` 3.14.1 `--rawvideo`** reproduced all five golden md5s exactly — giving C-oracle *pixels*, not just hashes.
3. **`dav1d` 1.5.4** came out byte-identical to aomdec, while rav1d-safe differed from both. Port defect, not a spec question.
4. Bisected by gating in-loop filters identically on both sides: deblock-only byte-identical to dav1d, CDEF-on diverged. The repo's own `__simd_test_log` comparator then named it — 447 CDEF SIMD-vs-scalar mismatches.

## Envelope — the issue's framing was wrong

#446 said "12-bit **inter**-frame". It is **10-bit and 12-bit, intra and inter**; 4 of 5 frames in the reproducer diverged, not 1. The inter framing was coincidental.

## Evidence

- All 5 frames now match the golden including `9f2291bf…`; all 8 animated tracks byte-identical to aomdec.
- Corpus vs dav1d: **10-bit 5/143 → 97/143**. 8-bit result file **byte-identical** before/after — provably inert where it should be.
- Full suite, independently re-run by the coordinator with one command on each side: **34 → 33 failing tests, ZERO new failures**, one repaired (`test_md5_verify_10bit_quantizer`); mismatch triples 1260 → 1182, 270 vectors changed hash, all toward the oracle.
- New gate pinned to the **oracle's** hashes and liveness-checked: reverting the fix fails it with exactly the pre-fix hashes.

## The bigger finding

The pre-existing failures are **one class**: aarch64 `safe_simd` kernels drifted from the generic scalar path, and this repo already ships the comparator that names each one.

- The remaining high-bit-depth failures are **exactly** the 91 vectors logging `MC_PREP_MISMATCH` (set equality; zero overlap with the 99 that pass) — sub-pel 8-tap `prep`, live in normal builds.
- 103 of 120 scanned 8-bit failures log `ITX_MISMATCH`.

Both left alone here as out of scope, documented in the commit. That is a follow-up worth more than this fix.

## Deliberately not fixed

The ARM CDEF padding sentinel (`8*bitdepth_max+1`, signed min *and* max) can win the `max` clamp where dav1d fills `0x8000` and uses `umin`/`smax` to exclude it. Real by inspection, but **zero** CDEF mismatches across 310 corpus vectors post-fix — no failing case, so no speculative fix.


lilith:
## Closing: the fix is on `main` and its regression vector now is too (#517, merged `fcf07c5`)

Triage of the 15 open PRs, 2026-08-13. This one was the only correctness item in the set, and it
was in the worst possible state: **the fix had landed and the gate had not.**

### Where the fix lives

`main` derives `pri_tap` in two places in `src/safe_simd/cdef_arm.rs` after the CDEF refactors that
followed this branch, and **both shift**:

* `filter_params():639` — `pri_tap: (4 - (pri_strength >> bitdepth_min_8 & 1)) as i16`, feeding
  `cdef_filter_block_16bpc_neon` through `..._dispatch`. This is the one that executes.
* `cdef_filter_block_scalar():968` — same expression, in the fallback whose own doc comment says it
  is "reachable only if `Arm64::summon()` fails, which it cannot on a NEON-mandatory aarch64 target".

This PR's diff touched `cdef_filter_block_16bpc_inner`, which the refactor split into those two, so
the fix carried through correctly. Verified by reading `main`, not by trusting the history.

### Where the gate lives now — #517, merged as `fcf07c5`

`tests/crash_vectors/arm_cdef_12bpc_422_pri_tap.obu` (245 B) + the per-frame oracle test in
`tests/decode_md5_committed.rs`. `git merge-base --is-ancestor 10773cd origin/main` passes; the
vector reads back off `origin/main`. CI: **23 of 23 checks green** (the 13 `disjoint-mut-ci.yml`
legs including both Miri legs are path-filtered to `crates/rav1d-disjoint-mut/**` and cannot fire on
a test-only branch — not applicable rather than skipped).

Teeth were proved rather than asserted, since a green gate for a wrong-pixels bug that cannot see
the bug is worse than no gate:

| site | mutation | test |
|---|---|---|
| `filter_params:639` | `>> bitdepth_min_8` removed | **FAILS**, 4 of 5 shown frames — matching this PR's own "4 of 5 frames diverged" |
| `cdef_filter_block_scalar:968` | `>> bitdepth_min_8` removed | passes — inert |
| `cdef_filter_block_scalar:965` | `panic!` planted at entry | passes → the function is **never called** |

Failure text from the first row:

```
12bpc 4:2:2 decode diverged from the aomdec/dav1d oracle on aarch64 (issue #446):
frame 1: expected 9f2291bf3f75dd440bc1d64ae26e0ac8, got 5c69666dc939fc1dc2a458cb3b61e917
frame 2: expected d4c7026d6bf22d5b973d59adbe73043d, got 95a1a33d2acce7339b05aa40e54e3321
frame 3: expected 484033dc917f8c49aba051070733e3ce, got fae265054b46b89dead0387e584863ab
frame 4: expected af113a069ef6a383e80522ce50943a99, got 6fbc7b83c599efb9909110329191736b
```

A liveness assertion was added on top of the version here: every shown frame must be **12bpc /
I422 / 64x64** before hashing. 8-bit is immune to this defect by construction
(`bitdepth_min_8 == 0`), so a later re-minimisation producing an 8-bit or 4:2:0 stream would have
left the MD5s checked and #446 ungated.

### What did NOT land, and where it is

`examples/obu_yuv_dump.rs` (277 lines) is code, not a record, and a records triage is the wrong
vehicle for it. It stays reachable at `pull/448/head` = **`6da3a29e7633921d2f1c496a4046612bcfe19362`**
(`git fetch origin pull/448/head`). The landed test is self-contained and does not need it.

Closing as superseded: the fix landed earlier through another branch, the gate landed in #517.
Nothing here is lost.

PR #450, opened 2026-08-07
perf(tile-threading): undo the 4-8% decode regression #445 shipped, and beat the pre-#445 baseline

### LINKED ISSUE #449: CDEF tap-window overlap: immutable read crosses a neighbouring block's mutable range under tile threading (aarch64/Windows-ARM)
A remaining instance of the #445 class, on the **immutable** half that #445 explicitly did not unify.

## Repro

zenavif CI, `Test (windows-11-arm)`, decoding a 12-bit animated AVIF at rav1d-safe `a6a7e232` (post-#445):

```
thread 'rav1d-worker-2' panicked at src/safe_simd/cdef_arm.rs:124:38:
        overlapping DisjointMut:
 current:  & _[2078..2086]    at src/safe_simd/cdef_arm.rs:124:38
 existing: &mut _[1824..2080] at src/cdef_apply.rs:59:26
```

Then `rav1d worker thread panicked during shutdown`, and the decode fails.

## Why this is the same class, and why #445 did not catch it

#445 fixed the **mutable** side: `narrow_guard_mut` reserved whole strided spans including inter-row gaps, and 29 call sites moved to `block_mut`. Its PR notes said plainly that `with_pixel_guard_immut` still open-codes the immutable half of the same policy and that this was a real remaining duplication left for follow-up. This is that follow-up arriving as a bug.

The overlap here is 2 bytes at a boundary (2078..2080), which matches CDEF's tap margin: `cdef_arm.rs` reads from `offset.wrapping_sub(2)`, i.e. two pixels left of its block, and lands inside the range the writer at `cdef_apply.rs:59` reserved for the neighbouring block. Under tile threading those are different workers on the same rows — genuinely concurrent, and genuinely disjoint in the pixels each actually touches.

## Why it matters more than a single panic

- It is on the **decode** path, from a bitstream our own encoder produces.
- `CpuLevel::Scalar` does not route around aarch64 CDEF (dispatch always returns `true` and consults no CPU mask), so there is no fallback.
- In `unchecked` builds the checker is absent, so the same over-broad reservation is simply not validated.

## Suggested shape

The same treatment #445 applied to the mutable side: give the immutable path a `BlockRef`-style counterpart so `with_pixel_guard_immut` and the direct `slice_as` readers share ONE policy, with the reader's tap margin reserved as part of its own window rather than colliding with the writer's. #445's PR already recommended this and named the reason it was deferred (it needs an immutable counterpart type, which would have widened a threading fix into an API addition).

## Related

- #445 (merged) — the mutable half.
- #448 (open) — a separate, genuine 16bpc CDEF `pri_tap` parity bug in the same file. Independent of this; #448 does not touch line 124.
- #447 — aarch64 NEON self-guided LRF index-out-of-bounds on 0.5.7.
- The aarch64 `safe_simd` drift class documented in #448 (91 `MC_PREP_MISMATCH` + 103 `ITX_MISMATCH` vectors).

Found by: imazen/zenavif PR #31 CI.

lilith:
**#445 shipped a measurable regression and I did not check for it.** This finds it, explains it, and fixes it.

## The regression (after/before, medians of 9 interleaved reps)

| vector | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| 256x144 8bpc | 1.022 | **1.073** | **1.065** | **1.062** |
| 1024x576 8bpc | 1.007 | **1.076** | **1.082** | **1.078** |
| 3840x2160 8bpc | 1.010 | **1.043** | **1.045** | **1.048** |
| 1024x576 10bpc | 1.001 | 1.008 | 1.002 | 1.008 |
| 3840x2160 10bpc | 0.997 | 1.004 | 1.003 | 1.008 |

**threads=1 is unchanged** (0.997-1.022) — the control holds, since the compact path is gated on `tile_threading_active()`. **10-bit is unaffected** (#445's converted sites are 8bpc-hot). On real multi-tile content there is no baseline to regress against: pre-#445 cannot decode 8-tile 4K at t>=2 at all, which is what #445 existed to fix.

## Root cause is NOT the memcpys

A throwaway probe — keep both memcpys, take one contiguous guard — measured **identical to baseline**. The copying is free; the entire cost is **2h extra tracker lock round-trips per block**. `sample` agrees: `BorrowTracker::add_mut` is ~13% of decode, second only to `decode_coefs`.

For the record, #445 did not invent this cost. At `0d350d1` the x86 `with_pixel_guard_mut` already did compact-read then work then compact-write-back under tile threading; #445 extended it to the aarch64/wasm safe-SIMD sites (itx_arm x12, mc_arm x7, mc_wasm x3, itx_wasm x2, cdef_wasm x2), which is where the 4-8% appeared on this box.

## The fix, and the one I rejected

Shipped: same per-row reservations, **registered 4 at a time under one lock** (`DisjointMut::copy_rows_out` / `copy_rows_in`). Every stored record stays a plain interval, so the hot scan is untouched.

Result vs pre-#445: 1024 8bpc **0.987-1.007**, 4K 8bpc **1.004-1.010**, 10-bit **0.978-1.001** — several cells now BEAT the pre-#445 baseline, because the loopfilter and ipred compact paths get batched too. Multi-tile 4K is 0.86-0.94x of main at t=2/4. Residual: 256x144 at t>=2 is still +4%.

**Rejected: one strided tracker record** — which would also have closed #449. Measured **worse twice** (+20% when the Bounds/slot arrays grew, +19% with a side table, because a strided borrow stays live almost continuously under tile threading). Discarded rather than shipped, with the reason recorded in the .meta so nobody retries it.

## Correctness

- **4,945 MISMATCH triples compared before/after: 0 differed.**
- Failing set **34 to 34, identical by name**. Did not grow.
- `mt_stress` passes at threads 1/2/4/8/16; the three ignored `tile_threading_overlap` tests also pass under `--ignored`. No test relaxed.
- Every sweep cell carries a frame md5; all arms, 7 vectors, 4 thread counts agree.

## Beyond the ask: tile threading ANTI-SCALES, and it predates #445

On 8-tile 4K: t=1 591 ms, t=2 471 (1.25x), t=4 578 (1.02x), **t=8 1101 ms, 1.85x SLOWER than single-threaded.** Same shape on the pre-#445 arm, so this is not new. Cause: every picture-plane borrow serialises on that plane's tracker spinlock.

`Settings::default()` is already `threads: 1`, so the default is safe. But the field's doc comment tells users they can set threads=0 "for better performance", and **measurement does not support that** on this hardware. Worth correcting, and worth treating the spinlock as the real scaling story.

## Harness notes

`ab_sweep.sh` now refuses to run alongside another instance or while the box is busy, and records load per cell — an orphaned background sweep and an editor `cargo check` at 844% CPU each silently invalidated a full dataset before I caught it.


lilith:
x86-64 datapoint (i265, Ultra 7 265K 20C, idle; `benchmarks/x64_i265_gap_2026-08-08.meta` + follow-up sweeps): the #445 per-row-guard regression this issue targets **does not reproduce on x86-64** — main/pre-445 is 0.97–1.06 across all 28 (vector × threads) cells (7 vectors, t=1..8, interleaved+rotated, medians of 9), and the `perf/audit-445` batching arm is likewise within noise of main there. The 4.3–8.2% cost is aarch64-specific in current measurements. Post-#454 main also supersedes part of this via the adaptive block shift. Suggest re-measuring the aarch64 delta on current main before deciding whether the batching lands.

lilith:
Correction to my x64 datapoint above: the `perf/audit-445` sentence is **void** — that benchmark arm was contaminated (a silent checkout refusal meant it rebuilt pre-merge main; details in imazen/rav1d-safe#458). What stands from genuine arms: the #445 regression itself (main vs pre-445, both genuine) does not reproduce on x86-64 (0.97–1.06 across 28 cells). The batching arm has **no valid x64 measurement**.

lilith:
## Closing: records landed in #518; the CODE is moot because #505/#513 replaced the population it batched

Triage of the 15 open PRs, 2026-08-13.

### The records landed, because `main` was dangling three citations at them

`benchmarks/tile_threading_guard_2026-08-06.meta` is referenced **by name** from
`shard_tracker_2026-08-07.meta:274`, `tracker_decomp_2026-08-07.meta:251` and
`x64_i265_gap_2026-08-08.meta:33` — the last two for its **vector recipe**. `main` was telling
readers to go consult a file it did not have. The meta and its three `.tsv.zst` are now on `main`
(#518, merged `71a0fbd`; `git merge-base --is-ancestor c942ffd origin/main` passes).

That preserves three things that existed nowhere else: the #445 regression per cell (8bpc t=2/4/8 at
1.043-1.082 with t=1 flat at 0.997-1.022 as the control, 10-bit unaffected), the probe that
**acquits the memcpys** (keep both memcpys, take one contiguous guard → identical to baseline, so
the cost was `2h` lock round-trips per block), and the vector recipe.

### Why the fix itself is not landed

`copy_rows_out` / `copy_rows_in` batched **four per-row registrations under one lock**, leaving each
stored record a plain interval so the hot scan was untouched. That was the right move against the
population as it stood. That population no longer stands: **#505 replaced the per-row registrations
with ONE exact strided-rectangle record and #513 made it the DEFAULT path on 2026-08-11**
(`Cargo.toml` now declares `__lf_rect` / `__rows_rect` as retained no-ops saying exactly that). There
is no longer a run of `h` per-row acquisitions at those sites for a batch of 4 to amortise.

### The dated supersede added over Result 3

Your §"Result 3 — a REJECTED attempt, recorded so it is not retried" is annotated on `main` rather
than rewritten, because it is **half right in a way that matters**. Its two designs are still
correctly refuted — do not put row geometry in `Bounds` (+20% at t=1 before the new path was even
reached), do not add an indirection to the dense scan (+19%). But it ends:

> Any future strided-record design has to leave the dense scan byte-identical, which is the
> constraint the shipped fix satisfies by not changing the scan at all.

**A design satisfying exactly that constraint was built five days later and it paid** — −2.0% to
−2.5% wall at t=8 on the 1024-wide family from the CDEF seam alone, −3.9% to −4.5% composed with
`LfBlock::fill`'s (`docs/LAYOUT_CDEF.md` §4). So the section now reads "recorded so it is not
retried **this way**", with the later result attached. Two further corrections are noted there: cost
tracks distinct shard lines visited rather than records filed (#505), and the "1.19-1.27" t=1 tail
sits inside the ±1.5% code-placement lottery #506 measured, though the "+20%" is far outside it.

### Also worth keeping from this round

Its §"Beyond the ask" reported that tile threading **anti-scales and it predates #445** (8-tile 4K:
t=1 591 ms, t=8 1101 ms, 1.85x slower than single-threaded) and named the cause as the plane
tracker's spinlock. That framing turned out to be the campaign's central thread — #499 later put
**100% of the added-CPU half of the tiled deficit on the borrow tracker** (`docs/TILED_SCALING.md`),
though #504 corrected the *lock* attribution specifically: contention is 0.264% of acquisitions on
`c256x2048` t=8, `lock_slow` is 10.7% of the tracker, and the cost is **coherence**, not waiting.

Reachable at `pull/450/head` = **`b5211a034d75fa3634c3f985e482e1122d807470`**; indexed in

PR #451, opened 2026-08-07
perf(tracker): kill the t=4 -> t=8 inversion — 1.40x on the best shippable config

lilith:
## What was wrong

On `perf/p2-kernels`, 8bpc `v4k_8tile` decoded **175.8 ms/frame at t=4 and 187.0 at t=8** — slower with twice the threads, so the best configuration was 4 threads. 10bpc did not invert.

## What it actually was

Diagnosed with three instruments, not guessed (full record: `benchmarks/p3_inversion_2026-08-07.meta`).

* **`probe-tasktime`:** occupancy is fine (7.55 of 8 workers busy at t=8). The *work* inflates — the same 136 tile-recon + 3x34 filter tasks cost 932 ms of CPU at t=4, 1511 at t=8, 2084 at t=16.
* **`sample`, waits excluded:** at t=8 that inflation is `TinyLock::lock_slow` 23.2%, `BorrowTracker::add` 24.3%, guard drop glue 10.2% — against `decode_coefs` at 16.8%.
* **`probe-untracked`** (tracker compiled out) does not invert at all: 336.5 / 95.5 / 57.7 / 65.2 ms at t=1/4/8/16. **The inversion is the borrow tracker, and at t=8 the tracker is ~70% of our wall.**

More shards fixes it, but cost 522.5 vs 412.5 ms at t=1 — and that cost is **not** the bigger table's cache footprint: a 128-shard array with only 32 shards reachable measures 528.0, i.e. no better. It is `add_wide`/`remove_wide` locking all `N_SHARDS`, and wide borrows are common exactly at t=1, because with tile threading off `block_mut` reserves the whole strided span.

## The change

* The wide path holds `0..=mask` instead of the whole array. `shard_of` ends in `& self.mask`, so every record and every narrow lock is inside that prefix. For a sub-64-KiB instance that is 1 lock instead of 128.
* `set_parallelism(n_tc)` from `rav1d_open`, beside `set_tile_threading` and monotone for the same reason: 32 active shards serial, 128 threaded.

## Measured (M4 Pro, idle-box guard, median of 9, no `nice` on a timed run)

| ms/frame | 8bpc t=1 | t=4 | t=8 | t=16 | 10bpc t=8 |
|---|---|---|---|---|---|
| head | 413.3 | 175.8 | 187.0 | 208.8 | 213.7 |
| this PR | 430.5 | **139.7** | **125.3** | **135.6** | **161.7** |

**Best shippable configuration 175.8 -> 125.3 ms, 1.40x**, and t=8 is the best thread count again. Gap to dav1d 1.5.4 at 8 threads: **5.23x -> 3.44x** (8bpc), 5.65x -> 4.18x (10bpc). Cost: **+4.2% at t=1**, which is the larger array itself and is not recovered.

A `--features shards-32` control arm reproduces `head` at every cell, so the win is the shard count and nothing else in the diff.

## Correctness

* Full corpus: 4,945 `MISMATCH` lines / 2,845 distinct triples — the exact counts on record for this lineage. Failing set is the documented 34.
* 288/288 campaign invocations bit-identical; all six local vectors identical to `head` at t=1 and t=8.
* `tests/wide_exclusion.rs` gates the wide path's exclusion with a **race**, and it is live in both directions: it fails on a deliberately shortened prefix (5 witnesses at `..1`, 2 at `..half`) and passes 20/20 here, plus 0/24 under 6-way process pressure.
* **Honest negative:** the single-threaded unit tests do *not* catch either break — they exercise the `state`/side-list half only. That is why the race gate is a separate file.
* `mt_stress_4k` 1/2/4/8/16, `tile_threading_parity` 3/3, `tile_threading_overlap` 3/3, concurrent-overlap stress 0/24. `forbid(unsafe_code)` holds, no `unsafe` added.

## Measured negative worth recording

**`TinyLock` spin backoff — the brief's leading hypothesis — is inside noise again**: -0.4/-0.5/-0.3% at t=1/4/8 on top of this change (the P1 campaign found the same on the 32-shard base). Kept as `--features probe-lock-backoff`, not enabled. The spin share was high because the *uncontended* fast path had itself tripled — cache-line ownership traffic, which a waiter's politeness does not touch.

## Still open (in the meta, by name)

The +4.2% single-thread residual; the tracker still being ~70% of the t=8 wall (125.3 vs 57.7 untracked) — the cure is fewer borrows, not more shards; and P2's `itx8`-slower-at-t=8 finding, which is now *explained* (`block_mut`'s compact path takes 2h guards where the scalar path takes h) but not fixed.


lilith:
## Closing: this already landed, and `main` has carried it since 2026-08-07

Triage of the 15 open PRs, 2026-08-13. Verified by SHA, not by title:

```
$ git merge-base --is-ancestor 761366e849a939c51fca2b001e10b2a25231dd45 origin/main
$ echo $?
0
```

The head commit `761366e8` **is an ancestor of `origin/main`**. Walking the ancestry path, the first
`Merge pull request` commit containing it is **#454** (`490ed78`, `verify/compose-2`) — this branch
was based on `perf/p2-kernels`, not `main`, and rode in through the compose chain rather than being
merged under its own number. `benchmarks/p3_inversion_2026-08-07.meta` is in-tree on `main` today.

So this PR has been a no-op since 2026-08-07 and there is nothing to merge, revert or preserve.

Indexed with the rest of the round at `benchmarks/pr_triage_2026-08-13.meta` (landed in #518).

PR #460, opened 2026-08-08
bench(scaling): the t1->t8 deficit is work inflation, not idleness

lilith:
Investigates the t=4 / t=8 scaling deficit (5.15x vs dav1d's 7.14x). **No performance change ships** — this is instruments, records, three measured negatives and one correction to the campaign's target number. Default build behaviour is untouched; corpus stays 766/766 by name.

## The standing suspect was wrong

The post-tile filter chain was the leading hypothesis. Measured on current `main`: per-worker busy time at t=8 is even to ±0.7% (62.5–63.4 ms/frame across 8 workers) and achieved occupancy is **7.143 of 8**. The pool is not starved.

Deficit at t=8 vs perfect 8x from t=1 (367.14/8 = 45.89 → measured 71.34), 25.45 ms/frame:

| term | ms/frame | share |
|---|---|---|
| CPU work inflation | 17.10 | 67% |
| lost occupancy | 7.57 | 30% |

**If CPU work did not inflate, the measured 7.143/8 occupancy alone would give 7.14x.**

## What it actually is

Total CPU per frame grows **+14.7% from t=2 to t=8** on a fixed 136 tile tasks and 34 filter sbrows — per-operation cost, not operation count. Wall-only interleaved A/B (n=5, idle box, ranges non-overlapping at every cell):

| t | plain | untracked | ratio |
|---|---|---|---|
| 1 | 364.66 | 287.17 | 0.7875 |
| 2 | 217.35 | 158.34 | 0.7285 |
| 4 | 116.66 | 82.63 | 0.7083 |
| 8 | 71.83 | 46.70 | **0.6501** |

scaling t1→t8: plain **5.077x**, untracked **6.149x**.

`sample` self-time names the riser: `TinyLock::lock_slow`, below the 0.38% cutoff at t=2 and **3.47% of non-idle at t=8**.

The collision is geometric. The adaptive shift gives a 4K plane shift 14 = ~4.3 picture rows per block (the crate's own doc); `v4k_8tile` is 4 tile columns, so all four concurrent workers in a tile row sit in one block and one shard. That is also why **t=4 → t=8 is the weak step**: 2 workers per tile row become 4.

## Measured negatives — do not repeat

- **Finer blocks to separate the columns: catastrophic.** t=8 ratio bs13 1.20, bs10 1.89, bs8 1.91; scaling collapses to 2.7x. The strided `block_mut` hull explodes. The block shift is a local optimum.
- **256 shards** (new default-off `shards-256` rung): worse everywhere — t=1 1.0730, t=4 1.0594, t=8 1.0295. Shard count is a local optimum too.
- **The wide path is not involved at all**: `probe-wide` reads `wide_total = 0` at t=1, 2, 4 *and* 8. Any plan premised on wide promotions at t=8 starts from a false premise.

## Correction to the target

7.14x is dav1d's **default**, which frame-threads; our checked build hard-pins `n_fc=1`. Measured in one interleaved sweep: dav1d `--framedelay 1` (the model we actually implement) scales **6.83x**, its default 7.24x. The like-for-like target is 6.83x. Small correction, but the figure in #455 is not the comparable one.

## New instruments

- `scripts/perf/scaling_ab.sh` + `scaling_report.py` — rotating-order interleaved multi-arm A/B across thread counts; prints min/max beside the median so a sub-3% claim can be checked against its own noise band.
- `probe_tasktime` `tailconc` / `tail_frac_of_wall` — concurrency histogram restricted to samples with no worker in a tile stage. Locates the occupancy loss: the post-tile tail is **0.61% of wall at t=2, 1.08% at t=4, 11.96% at t=8**. Its own `TILE_ACTIVE` counter is contended at t=8 and inflates that build's absolute stage ms; the fractions are sample-count ratios and survive. Caveat documented at the probe and in the meta.

## Gates

766/766, set-diff **by name** with the md5 inside the key against `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`: 0 differing. `mt_stress` 1/2/4/8/16 ok. `multi_decoder_pressure` 12 concurrent decoders PASS against serial reference md5s. `tile_threading_overlap` / `reproduce_overlap` / `reproduce_overlap_scalar` pass under `--ignored`. disjoint-mut `soundness` (25), `shard_liveness` (5), `narrow_release`, `wide_exclusion` green.

**Teeth proved by mutation, not by green**: deleting `add()`'s in-lock `state` re-read makes `wide_exclusion` FAIL; restoring it makes it pass.

## Not done

The one principled fix left for the tracker contention is #455's untested item: stop reserving the strided hull in `block_mut`/`narrow_guard` and register per row-group. That is what would let the block shift go fine enough to separate tile columns without the blowup above. It is soundness-critical and adjacent to the March-2026 strided tracker that was merged and reverted the next day — not attempted here, rather than attempted and left unproven.


lilith:
Verified independently and **merged into `main` via #462** (composed with #461). `2bfe1e6` and `5d491f3` are both ancestors of `main@origin`.

Two of your load-bearing claims reproduce independently on my own sweep (n=7 for the gap, n=5 for the tracker A/B; idle box, foreign=0 on every committed row):

- **The 6.83x target correction.** dav1d 1.5.4 `--framedelay 1` scales t1→t8 at **6.84x** here (your 6.83x), its default at **7.11x** (your 7.24x). Frame threading is worth 4% to dav1d at t=8 (34.7 vs 36.1 ms). The correction stands.
- **The tracker is the inflation.** `untracked/base` = **0.778** at 8bpc t=1 and **0.649** at t=8 (yours: 0.7875 / 0.6501), and removing it takes scaling 5.02x → 6.02x (yours 5.08 → 6.15).

One correction to a number you and the index both quote: **base t1→t8 is 4.93x on the two-point wall fit**, not 5.15x. 5.15x is the in-process instrument (`scaling_ab.sh`); mine is `verify_gap.sh`. Both honest, and the t=8 cell is the noisy end (base band 70.7..75.9 at n=7) — but the figure should travel with its instrument, because 4.93 vs 5.15 is a 4.5% spread on the campaign's headline scaling number.

Composed with #461, scaling is unchanged (4.93x → 4.95x), exactly as you said it would be.

I did NOT re-run your probe-tasktime occupancy figures, the blockshift ladder or the shards-256 rung. Your `shards-256` rung composes clean and is correctly excluded from the default `N_SHARDS` cfg chain.

Also confirmed from your report: `test-vectors/bench/photo_4k.avif` is gitignored and `mt_stress` hard-fails in a fresh worktree. And your `style: cargo fmt` split-out is load-bearing — `cargo fmt --all --check` is **red on main** at two sites, so composing it turns the `Format` CI job green.

PR #461, opened 2026-08-08
perf(tracker): the borrow path is latency-bound, not throughput-bound — 1.47x -> 1.37x at 8bpc t=1 (#455)

lilith:
Closes the "cut the per-borrow cost of the tracker" item from #455. The bar (1.30x at 8bpc t=1, i.e. 320 ms) is **NOT met** — this lands 337.1 ms = 1.37x, 17 ms short, with 46.9 ms/frame of tracker still on that cell.

## What the run actually established

Three measurements, each refuting the previous plan:

1. `examples/probe_borrow_cost` (new) prices one acquire+release pair at 7.89 ns, the shipped three-RMW sequence at 3.50 ns against 1.54 for one RMW plus plain stores. At the measured 15.65 M pairs/frame (`probe-count`), deleting two of three RMWs predicted ~26 ms/frame.
2. Deleting them measured **-0.6%** at 8bpc t=1. Removing 26 more instructions (the whole 0x80 stack frame) measured **null** on the same cell.
3. So `probe-addnop` (new throwaway feature: keep the call, delete the body) was built to settle it — **290.97 ms against 365.77 real and 287.11 with no tracker**. The call barrier is 4.4% of the tracker's cost; 95.6% is `add`'s own latency.

**This path is latency-bound. Instruction-count and atomic-count fixes do not pay; dependency-chain fixes do.** Everything that shipped after that point shortens the chain, and every one paid.

## Shipped (ratio vs base at 8bpc / 10bpc t=1, each step its own interleaved sweep, median of 9, idle box)

| change | 8bpc | 10bpc |
|---|---|---|
| per-slot live flags (publish/release become plain stores, 2 locked RMWs gone) | .9894 | .9786 |
| + frameless leaf `add` (try_lock + cold outlined contended path; scalar `overlap_panic`) | .9844 | .9720 |
| + `mask == 0` skips the shift load, both block divisions and the hash | .9553 | .9458 |
| + `live_mask` straight-line when `allocated <= 1` | .9455 | .9397 |
| + `alloc` straight-line on an empty shard; `allocated` written once | .9348 | .9211 |
| + drop a sub-branch; store `mutable` whole instead of RMW | **.9310** | **.9201** |

All 8 cells, median of 7: 8bpc t=1/2/4/8 `.931 .912 .920 .924`; 10bpc `.920 .916 .923 .913`.

## Gap to dav1d 1.5.4 (same interleaved sweep, n=5, idle, foreign=0)

| vector | t | base | head | dav1d | base | head |
|---|---|---|---|---|---|---|
| v4k_8tile | 1 | 362.5 | 337.1 | 246.4 | 1.47x | **1.37x** |
| v4k_8tile | 2 | 217.1 | 197.1 | 125.2 | 1.73x | **1.57x** |
| v4k_8tile | 4 | 117.1 | 107.4 | 65.6 | 1.78x | **1.64x** |
| v4k_8tile | 8 | 74.7 | 67.3 | 36.2 | 2.07x | **1.86x** |
| v4k_8tile_10b | 1 | 414.9 | 382.4 | 251.3 | 1.65x | **1.52x** |
| v4k_8tile_10b | 8 | 77.7 | 71.8 | 37.2 | 2.09x | **1.93x** |

## Reverted, with numbers

- `fold` (duplicate the body so shard 0's address constant-folds): **+0.9%** — 117 -> 208 instructions and the frame comes back.
- `findeo` (`find` early-out on `occupied == 0`): **+2.3%** — the load is free and LLVM already folds the test into one `ands`+`b.eq`.
- `chain3` (also drop the pre-lock `state` check): **+0.8%** — that load warms the header line while the acquire is in flight.
- Micro-only: fusing lock+occupancy into one word 4.76 ns vs 3.50 (three RMWs on one address serialise); Relaxed publish 4.82 vs 3.50.

## Candidate 4 is REFUTED, not untested

`probe-wide` on the shipped build: t=1 `slow=0 multi=0 wide_total=0` — `mask == 0` takes the strided hull whole and `find` is O(live slots), independent of extent, so row-groups could only ADD registrations. t=8: 17,275 of ~15.65 M adds = **0.11%**.

## Candidate 3 costed, not built

Publish + `dmb ish` + peer scan + release = **0.599 ns/pair** vs 3.50 shipped; `dmb ish` alone 0.246 ns. Blocked on the Dekker hazard at thread join — needs an RCU-style grace period, and the scan is O(registrants) over lines they write (all 8 workers touch all 12 hot planes). Full write-up in the `.meta`.

## Soundness

Publish and release are plain stores because a per-slot flag byte has **at most one writer**: the allocator, which holds the lock AND has observed the byte zero, and that borrow's owner, once. `ShardRecs.allocated` is a lock-protected SUPERSET so the scan stays one branch wide.

Four hazards planted on the exact shipping code, each confirmed to FAIL under `--features __probe_wide`, each restored with a 0-line diff:
- (a) delete `add`'s in-lock `state` re-read (the 4af62ae TOCTOU) — 1206 missed overlaps
- (b) `active()` cut to one shard — 3 missed overlaps
- (c) a published slot never enters `allocated` (new invariant)
- (d) `allocated` narrowed to 0, dropping live slots (new invariant)

`forbid(unsafe_code)` proved actively (planted `unsafe` fails the default build). **766/766** corpus, `md5_setdiff` CLEAN **by name** against `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`. `mt_stress` 1/2/4/8/16. `multi_decoder_pressure.sh` 12 procs x 3 iters x {1,2,4,8,16} threads, all md5s match the serial reference. Every arm x every cell produced identical plane md5s.

Full record: `benchmarks/tracker_borrowcost_2026-08-08.{meta,tsv}` + `_gap_` + `_micro_`.


lilith:
Verified independently and **merged into `main` via #462** (compose with #460). Every commit here (`1a2c165`, `d48c240`, `d1d5408`) is now an ancestor of `main@origin` — checked with `git merge-base --is-ancestor`.

Re-measured on my own build of `main` as the base, n=7, idle box, foreign=0 on all 224 rows, dav1d 1.5.4 in the same interleaved sweep: **337.3 ms and 1.368x at 8bpc t=1** against your 337.1 / 1.37x. Bar met nowhere; 16.7 ms short at the best cell. Full table, band check, decomposition and mutation battery in #462 and `benchmarks/verify_compose5_2026-08-08.meta`.

Three additions worth carrying forward:

- **Your `add_contended` is genuinely gated, and you undersold it.** You deleted the in-lock `state` re-read at *both* sites at once, which cannot distinguish the new cold path from the old fast one. Deleting it in `add_contended` ALONE fails `wide_exclusion` 5/5 with 1555–2960 missed overlaps — two orders of magnitude more than the fast-path-only arm's 8. I proved the path is reached (a `loop{}` there hangs the test) rather than assuming it.
- **`addnop` vs `untracked` prices the call barrier at 3.8%**, not 4.4% — same conclusion, and `untracked/base` lands at 0.778 / 0.649 against your .7913 / .6501.
- **The attribution table is t=1-only and the headline mechanism cannot fire at t>1.** `src/lib.rs:181` calls `set_parallelism(n_tc)`; `n_tc > 1` raises `ACTIVE_SHARDS` to 128, so `mask_for` returns 127 on every picture plane and the `mask == 0` skip only reaches the ~1,027 sub-64-KiB instances. The measured t>1 wins (0.910–0.926) are real and are bought by the other four mechanisms. Nothing to fix — just don't quote 'this is the big one' next to the multi-thread cells.

Closing as merged.

PR #466, opened 2026-08-08
fix(soundness): decode_permutations was a dead gate on aarch64 (16/19 failing) — gate on summon(), run it on aarch64

lilith:
## The defect

`tests/decode_permutations.rs` is the **only** gate on "every archmage token
permutation produces bit-identical output". On unmodified `main` @ 8568dbb it
**fails 16 of 19 tests on aarch64** under the exact CI invocation:

```
cargo nextest run --profile ci --no-default-features \
  --features "bitdepth_8,bitdepth_16" --release --test decode_permutations
```

```
thread '...' panicked at src/safe_simd/mc_arm.rs:5931:37:
called `Option::unwrap()` on a `None` value
```

CI reported green throughout, because `.github/workflows/ci.yml:245` pinned the
job to `ubuntu-latest` — where every `src/safe_simd/*_arm.rs` file is compiled
out. The architecture this repo has spent a week porting NEON kernels to is the
one architecture the parity gate never ran on. `--test-threads=1` does not help;
the file's old header blaming parallelism was a red herring for this failure.

Second dead gate found in two rounds (the first: `wide_exclusion` went vacuous
when a config change routed its instance away from the wide path).

## Root cause

Nine `Arm64::summon().unwrap()` calls *inside* dispatch bodies. The permutation
harness disables the NEON token **process-wide on purpose**; these dispatchers
asserted it instead of falling back.

| file | sites |
|---|---|
| `src/safe_simd/mc_arm.rs` | 6 — `avg`/`w_avg`/`mask`/`w_mask`/`mc_put` `_inner`, `mct_prep_dispatch` |
| `src/safe_simd/filmgrain_arm.rs` | 3 — `fgy_inner_8bpc`, `fgy_inner_16bpc`, `fguv_inner_8bpc` |

## Fix

1. **Gate, don't assert.** mc_arm uses
   `let Some(token) = Arm64::summon() else { return false }` — the shape
   `loopfilter_arm::lpf_dispatch` and `looprestoration_arm::lr_filter_dispatch`
   already use; `false` routes the block to the scalar reference in `src/mc.rs`,
   which every caller already handles. filmgrain_arm has no bool seam at that
   depth, so it holds an `Option<Arm64>` and runs a scalar row twin — those
   loops already existed as `cfg(not(target_arch = "aarch64"))` blocks, dead
   code inside an aarch64-only module, and are now the live fallback.
2. **Run the gate where it can fail.** `conformance` and `permutation-tests`
   both gain an `ubuntu-24.04-arm` (aarch64) leg. aarch64 conformance gets 60
   minutes: the scalar tier now really decodes the whole 8-bit/data corpus
   instead of panicking out in milliseconds.
3. **The gate asserts its own liveness.** Fails when only one permutation ran
   (compile-time-guaranteed tokens make the comparison vacuous) and when every
   vector was skipped (missing corpus). Both were silent-green before.
4. **Whole-class audit.** Zero `summon().unwrap()` remain in `src/`. Remaining
   `expect()`s are in `#[cfg(test)]` modules — all now hold `token_test_lock()`
   per `safe_simd/mod.rs`'s own documented rule — or in four **unreferenced**
   x86 `extern "C"` AVX2 loop-filter shims, documented in place. Their message
   "AVX2 implies Desktop64" was false (`Desktop64` is x86-64-v3, a strict
   superset of AVX2) and is corrected.

## Before / after on aarch64, BY NAME

Darwin 25.5.0, M-series, exact CI invocation.

**Before — 3 pass:** `test_permutations_smoke`, `test_permutations_8bit_intra`,
`test_permutations_10bit_features`.
**Before — 16 fail:** `8bit_data`, `8bit_features`, `8bit_issues`,
`8bit_quantizer`, `8bit_size`, `8bit_film_grain`, `8bit_cdfupdate`, `8bit_mfmv`,
`8bit_mv`, `8bit_resize`, `10bit_data`, `10bit_quantizer`, `10bit_issues`,
`10bit_film_grain`, `12bit_data`, `12bit_features`.
Panic sites: `mc_arm.rs:5931` ×14, `filmgrain_arm.rs:458`, `filmgrain_arm.rs:695`.

**After — 19/19 pass** (605 s wall).

## Teeth proven, not assumed

Plant → confirm FAIL → restore → confirm green, three times:

1. NEON kernel divergence (`fgy_row_neon_8bpc`, max clamp −1) → `8bit_film_grain` **FAILS**.
2. NEON kernel divergence (`put_8tap_8bpc_inner`, `intermediate_bits` 4→3) → `8bit_mv` + `8bit_mfmv` **FAIL**.
3. Scalar-fallback divergence (`fgy_inner_8bpc`, noise +1) → `decode_permutations` **FAILS** while
   `decode_md5_verify` stays **14/14 GREEN**. That is the important one: the fallback
   paths this PR makes live are a surface no other gate in the repo covers.

## Corpus unchanged

`md5_inventory`: **pass=766 mismatch=0 error=0**. Set-diff by *name + actual md5*
against `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst` is **EMPTY**
(768 rows incl. the 2 SKIPs).

Also on this tree: `decode_permutations` + `decode_md5_verify` 33/33;
`cargo nextest --lib` 69/69 and plain `cargo test --lib` 69/69 (the lock change
matters only for the latter); clippy safe-simd `-D warnings` clean;
`cargo fmt --all --check` clean. `clippy --features c-ffi` hits the pre-existing
macOS-only `E0080` `EAGAIN` const assert in `src/error.rs`, untouched here (d09565f).

Refs #455.


lilith:
## CI green — including the leg that could not previously fail

Run [31278238028](https://github.com/imazen/rav1d-safe/actions/runs/31278238028): **21/21 jobs success**.

| new job | result |
|---|---|
| `Conformance (decode permutations, aarch64)` | **19/19 passed, 1380 s** (23 min) |
| `Conformance (decode permutations, x86_64)` | 19/19 passed |
| `SIMD Permutation Tests (aarch64)` | passed |
| `SIMD Permutation Tests (x86_64)` | passed |

23 min is why the aarch64 conformance leg carries `timeout-minutes: 60` instead
of the x86 leg's 30 — the scalar tier now genuinely decodes the whole
8-bit/data corpus there, where before the fix it panicked out in milliseconds.

## Perf: null, measured

The mc_arm half is cost-identical by construction (`summon()` + a branch either
way). The filmgrain_arm half is not — the token became a loop-invariant
`Option<Arm64>` matched once per block-row — so it was A/B'd instead of
asserted. `profile_ivf` on `8-bit/film_grain/av1-1-b8-23-film_grain-50.ivf`
(`apply_grain` defaults on, so the changed path really runs), 300 in-process
iterations per rep, threads=1, no `target-cpu=native`, arms interleaved with the
order rotated every rep, n=9 per arm, base `8568dbb` vs head `9ef5f42`, box at
load ~2.9 (not idle — ratios are the claim, absolutes are not):

```
base  median 1.9126 ms/frame  min 1.8958  max 1.9880
head  median 1.9390 ms/frame  min 1.8937  max 1.9763
paired head/base: 1.0153 1.0015 1.0184 0.9920 1.0142 0.9889 1.0149 0.9821 0.9830
median paired ratio 1.0015x  (min 0.9821, max 1.0184)
```

Arm ranges overlap almost completely and the paired ratios straddle 1.0 in both
directions — **null**. Record committed at
`benchmarks/permgate_filmgrain_ab_2026-08-08.{tsv,meta}`.

lilith:
Independent verification (verify/compose-6, `ca570e7`). Record: `benchmarks/verify_compose6_2026-08-08.meta`.

**Everything in this PR checks out, and the dead-gate claim is real.** Branch roots at `main` @ 8568dbb (checked first).

On unmodified `main`, the exact CI invocation on aarch64 fails, panicking at precisely the sites you named — `mc_arm.rs:5931`, `filmgrain_arm.rs:458`, `filmgrain_arm.rs:695`. On the composed tree: **19/19 PASS, 575 s.**

I planted my own divergences rather than take yours on trust, and the two of them partition the permutation space exactly:

- **NEON kernel** (`fgy_row_neon_8bpc`, `max_val.wrapping_sub(1)`) → `8bit_film_grain` fails in **24 of 25** permutations; the base-NEON-off one is untouched, i.e. the fallback there produced the correct pixels.
- **The scalar twin this PR made live** (`+1` on the noise term in `fgy_inner_8bpc`'s `None` branch) → fails in **exactly the 1 remaining** permutation, *while `decode_md5_verify` stays 14/14 GREEN with the mutation still planted.*

24 + 1 = 25 = `permutations_run`. That is the whole case for this PR in one number: the fallback paths it makes live are a surface the 766-vector corpus gate structurally cannot see.

Liveness assertion also has teeth — raising the bound to 100000 reports "only 25 token permutation ran", so it is reachable and 25 ≥ 2 is not vacuous.

Corpus on the compose: 766 PASS / 0 mismatch / 0 error / 2 skip, set-diff BY NAME with the actual md5 in the value against the committed baseline — 768 rows each side, **EMPTY**.

Diff audit: every removed line is accounted for — 38 `cfg_attr(debug_assertions, track_caller)` widened (not dropped), 9 `Arm64::summon().unwrap()` → 6 `let Some(..) else` + 3 held `Option`, 4 `Desktop64` expects re-messaged, and the net movement is **+26 assertions**. Nothing silently deleted.

Perf: your film-grain null holds. My whole-decode sweep (n=7, dav1d in the same interleaved sweep) shows this branch is not where the movement is — that is #467's half.

One note, not a blocker and not yours to fix here: `cargo clippy --all-targets -- -D warnings` fails on aarch64 on `main` too, and beyond the `itx_arm*.rs` dead code you named it also hits `examples/md5_inventory.rs`, `examples/md5_ablate.rs`, `examples/bench_ivf_limit.rs` and `benches/tier_isolation.rs` — identical on both trees, so the CI Clippy leg being x86-pinned hides four more targets than the note says.

PR #467, opened 2026-08-08
perf: cut the NUMBER of borrow pairs — 8bpc t=1 1.368x -> 1.294x (#455)

lilith:
## 8bpc t=1 **1.368x -> 1.294x** — the ~1.30x bar is met at that cell, by cutting the NUMBER of borrow pairs

Follows #455 / #465. Record: `benchmarks/borrow_pairs_2026-08-08.meta` + three raw files.

### The map that was missing

`sample` attributes `add`/`remove` to whichever caller LLVM inlined them into, so it can say the tracker is 41.7 ms/frame over ~25 sites with no hot spot, and nothing more. A new throwaway per-call-site **counter** (`--features probe-sites`, `crates/rav1d-disjoint-mut/src/site_probe.rs`) says the distribution is the opposite of flat. `v4k_8tile` 8bpc t=1, registrations per frame:

| count | share | site | mean extent |
|---:|---:|---|---:|
| 3,835,042 | 24.5% | `loopfilter.rs` `LfBlock::fill` | 8.8 B |
| 2,534,988 | 16.2% | `ctx.rs` `CaseSetter::set_disjoint` | 2.1 B |
| 1,232,696 | 7.9% | `ipred_arm.rs` cfl_pred | 10.1 B |
| 736,520 | 4.7% | `ipred.rs` splat_dc | 10.4 B |
| 646,912 ×2 | 8.2% | `cdef_arm.rs` padding + block write | 10.0 / 8.0 B |

Top 3 = 48.6%, top 7 = 65.6%, 68 sites total. Every picture-plane entry is the same shape: a `w×h` **strided block touched one row at a time** — `h` registrations of ~10 bytes.

### The change

One hull registration instead of `h`, **only when `tile_threading_active()` is false**. That is not a new policy: `block_mut`, `with_pixel_guard_immut` and `compact_read` already switch on the same monotone latch, because the hull also reserves the inter-row gaps and AV1 tiles partition a frame by *columns* — under tile threading a neighbour's legitimate write lands in a reserved gap and `DisjointMut` reports a **false** overlap.

`Rav1dPictureDataComponentOffset::for_rows` / `for_rows_mut` is the shared implementation; `LfBlock::fill_hull` is open-coded because `fill` is handed an explicit stride. Converted: the loopfilter read, 11 `ipred` predictors, NEON cfl_pred, the NEON cdef padding/write/dir-search, and `backup2x8`. `pal_pred_rust` took one guard per **two pixels** and now takes one per row — a pure count cut with no extent change at any thread count.

**Detection is never weakened.** A hull is a superset of the `h` row ranges, so it conflicts with strictly more; it cannot miss an overlap the narrow ones would catch. Only false positives are at stake, and the latch rules those out.

15,646,727 -> 7,924,706 registrations/frame (-49.4%).

### Measured (n=7, interleaved, dav1d 1.5.4 in the same sweep, bands printed)

| | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| 8bpc base | 1.368 | 1.580 | 1.625 | 1.869 |
| **8bpc head** | **1.294** | 1.581 | 1.628 | 1.875 |
| 10bpc base | 1.520 | 1.616 | 1.666 | 1.935 |
| **10bpc head** | **1.437** | 1.622 | 1.667 | 1.925 |

8bpc t=1: 337.5 [334.2..342.2] -> 319.2 [318.0..321.8] ms/frame, **disjoint bands**. The base column reproduces #465's 1.368 to three digits.

**t>1 did not move and cannot** — the hull is gated off there. Stated as a control and proved by count, not timing: `probe-sites` reports **22,700,725 registrations per frame on both base and head at t=8**, same 68 sites.

### Two ceiling arms that redirect the next round

`tinynop` is a throwaway UNSOUND build that stops tracking sub-`SHARD_MIN_LEN` instances; `untracked` removes the tracker.

| cell | base | head | tinynop | untracked | dav1d | tracker residual | of which tiny / picture |
| t=1 | 338.9 | 318.1 | 299.4 | 286.3 | 246.6 | 31.8 | **18.7** / 13.1 |
| t=8 | 65.2 | 65.6 | 63.8 | 45.9 | 34.4 | 19.7 | 1.8 / **17.9** |

- At t=1 the residual is **59% the `BlockContext` class** (twenty separate 32-byte `DisjointMut`s), not the picture planes.
- At t=8 it inverts, and **even a zero-cost tracker leaves t=8 at 1.335x** — the bar is not reachable there by tracker work at all. The t=1 untracked ceiling is 1.161x.

### Soundness

New mechanism gate (`row_guard_policy_tests`): two tests asserting *which extent* got reserved. The threading-OFF half runs in a **child process** (the latch is monotone and another test in the same binary sets it) and checks a **marker line**, not the exit status — libtest exits 0 on an empty filter, which is exactly how #459's `wide_exclusion` went vacuous. Teeth proved by three mutations (`if true` / `if false` / renamed child), each restored from a sha256 snapshot.

Existing gates re-proved on this tree, both restored byte-exactly: the 4af62ae in-lock `state` re-read deleted -> `wide_exclusion` FAILS (4 witnesses); `active()` cut to one shard -> FAILS (13 witnesses). `#![forbid(unsafe_code)]` proved actively. `soundness` 25/25, `shard_liveness` 5/5, `narrow_release` 1/1.

Corpus **766/766**, set-diff BY NAME with the md5 in the value: 0 only-in-baseline, 0 only-in-head, **0 differing**. `mt_stress` 1/2/4/8/16, 12-decoder pressure PASS.

### Retraction inside this branch

`40f2a07` added `#[inline(always)]` on the strength of an **n=1** reading (head/base 1.0295 at 8bpc t=8). At n=7 with both forms as separate arms every t>1 band overlaps and plain `#[inline]` has the better median at 7 of 8 cells. `fcade5e` reverts it and says so. Same shape as the 2.7% claim #465 retracted; `scripts/perf/gap_bands.py` (added here) prints min/max and a disjoint/OVERLAP verdict so the next one is caught before it ships.

### Not done

x86_64 not measured. Only `v4k_8tile{,_10b}`. `ctx.rs` (32% of what remains, and the only thread-count-independent lever) untouched — no safe batching exists at the call site and the one-`DisjointMut`-`BlockContext` refactor needs a concurrency argument about `f.a[t.a]` this round did not establish. NEON `cfl_ac` (629k/frame) not converted: it needs two rows at once, which `for_rows`'s callback shape does not fit.

Separately found, pre-existing, **not** fixed: `narrow_guard`/`narrow_guard_mut`'s negative-stride branch computes `start = offset + 1 - total`, a span shifted `w-1` below the rows it stands for — i.e. narrower than the bytes it hands out. Unreachable in the safe build (positive strides only); `for_rows` deliberately does not copy the formula.


lilith:
Independent verification (verify/compose-6, `ca570e7`, pushed to `origin`). Full record: `benchmarks/verify_compose6_2026-08-08.meta` on that branch.

**Confirmed.** Composed onto `main` @ 8568dbb with #466 (both branches root at 8568dbb — checked first). Corpus 766/766, set-diff BY NAME with the actual md5 in the value against `aarch64_md5_fixes_2026-08-07_final.tsv.zst`: 768 rows each side, **EMPTY**. Re-measured the gap myself, n=7, dav1d 1.5.4 in the same interleaved sweep, two-point wall fit:

| cell | head/base | bands | base/dav1d | head/dav1d |
|---|---|---|---|---|
| 8bpc t=1 | 0.9382 | **disjoint** | 1.368 | **1.283** |
| 8bpc t=2/4/8 | 0.997 / 1.017 / 0.982 | OVERLAP | 1.584 / 1.731 / 1.945 | 1.579 / 1.760 / 1.909 |
| 10bpc t=1 | 0.9421 | **disjoint** | 1.512 | **1.425** |
| 10bpc t=2/4/8 | 1.013 / 1.055 / 1.007 | OVERLAP | 1.597 / 1.665 / 1.957 | 1.617 / 1.756 / 1.971 |

Your 1.368 → 1.294 reproduces (I get 1.283) on a differently-loaded box, and t>1 is null in both directions exactly as you reported. Your `row_guard_policy` mutation-G equivalent, and the two `__probe_wide` hazards, all fail-then-restore as claimed.

**One blocking defect, not in your report: this branch turns CI red.**

`Build & Test (ubuntu-latest, asm)` fails on `5580533` (run 31279664234). `main` and #466 are both fully green, so it is this branch. The new `row_guard_policy_tests` assert tracker behaviour, and `asm` implies `unchecked`, which compiles borrow tracking out — so both tests fail, one of them via its own anti-vacuity assertion ("tracking is off in this build and this test gates nothing"). The design intent is right; the cfg gate is missing.

Reproduced locally, isolating it exactly:

```
cargo test --release --no-default-features --features "bitdepth_8,bitdepth_16,unchecked" --lib
  main    69 passed; 0 failed
  compose 69 passed; 2 failed   <- only the two new tests
```

One attribute, using the idiom already in tree (`src/disjoint_mut.rs:29`, `src/safe_simd/pixel_access.rs:451`):

```diff
-#[cfg(test)]
+#[cfg(all(test, not(feature = "unchecked")))]
 mod row_guard_policy_tests {
```

**Checked and dismissed — do not chase these:** the four `disjoint-mut CI` failures on this PR (`Test --no-default-features` ×3, `Test --all-features`) are **pre-existing on `main`** — I reproduced them identically on both trees (23 and 2 errors). That workflow is path-filtered on `crates/rav1d-disjoint-mut/**`, so it just never fired on #466 or recent `main` pushes. Same for the aarch64 `clippy --all-targets` failures (dead code in `itx_arm*.rs` plus four example/bench targets) — identical on both trees.

**Where the ~1.30× bar stands:** met at 1 of 8 cells (8bpc t=1). Your `untracked` ceiling recomputes to 1.332× at t=8 from your own committed decomposition, so a zero-cost tracker still misses there — the t>1 residual is scaling (my t1→t8: head 4.34×, base 4.54×, dav1d 6.46×), not borrow cost. My bucket profile (`/usr/bin/sample` + `bucket_selftime.py`) puts the tracker at 43.3 → 30.1 ms/frame at 8bpc t=1, ≈63% of the 21.0 ms/frame wall win; the 43.3 agrees with #465's independently-measured 41.7.

PR #468, opened 2026-08-09
Design A (static recon row-split): blocked on ownership, realizable form already landed, largest sound subset measures null

lilith:
Measurement round: the data chose **Design A**. This round built it, and the honest result is that
A as specified cannot be written in safe Rust in this architecture, that its realizable form
already landed in #467, and that the largest sound subset left measures **null**.

Base `main` @ ee07b00 (the #467 merge) — **not** 8568dbb; see §2, that difference is half the
finding. Record: [`benchmarks/recon_rowsplit_2026-08-09.meta`](../blob/perf/recon-rowsplit/benchmarks/recon_rowsplit_2026-08-09.meta)
+ `recon_rowsplit_gap_2026-08-09.tsv` (216 rows) + `recon_rowsplit_sites_2026-08-09.tsv` (census).

**Not proposed as a perf win.** It is a real registration reduction with a documented null, a new
sound helper, a new mechanism gate, and two entries in the brief. Merge it for those or close it —
either is a fair reading.

### 1. Design A as specified is not expressible here — STOP, not `unsafe`

The design's premise is a `&mut` to the picture plane at tile setup to `chunks_mut`. There is none,
at any thread count:

- `include/dav1d/picture.rs:1287` — planes live behind `Arc<Rav1dPictureData>`, shared with
  `sr_cur`, the ref slots and the output queue while the frame decodes.
- `src/thread_task.rs:1237` (and `:1416`, `:1436`, `:1451`) — every tile **and** filter task opens
  `fc.data.try_read()` and gets `&Rav1dFrameData`. Shared read guards *are* the tile parallelism.
- `src/lib.rs:221` — workers are `'static` threads spawned once against a global task queue, so no
  scoped-thread lifetime can carry a split into them.

So a "static split" would mean handing out `&mut` on a hand-written invariant instead of a compiler
proof. Per the task's soundness instruction: stopped and reported, not written.

Worth noting the blocker is **upstream of the expected one**. The anticipated hard part was recon's
access pattern — overlapping simultaneous views. That part is fine: `ipred_prepare` materialises the
neighbour edge into scratch first, so the prediction kernels are pure block writers, and intra reads
do not cross a tile column (`src/ipred_prepare.rs:127-131`, `:229`). The site-class round checked
that premise and reported it holding; it did not check `&mut` availability.

### 2. A's realizable form already landed, 44 minutes after the decision's base

There *is* a form that works: one tracked guard over the strided hull, then let borrowck split that
into rows. That is #467's `for_rows{,_mut}` — merged as ee07b00 at 16:44, **after** the site-class
base 8568dbb — and it is already on every top `recon+big` site the A-vs-B decision was computed on.

It is gated off above one thread and necessarily so: the hull reserves inter-row gaps belonging to
other tile COLUMNS (#467 measured 8-9 spurious panics in 24 concurrent runs). Verified here — the
t=8 census is **22,700,725 registrations on both 8568dbb and ee07b00, bit-for-bit unmoved**.

**So the property A was chosen for — a share that holds across thread counts — is not available to
any realizable form of A in this architecture.** Every form that can be written is t=1-only, which
is B's coverage. The A-vs-B choice was sound on the numbers and is moot on the architecture.

### 3. What this builds: the one picture site #467 could not convert

#467 §5c left exactly one `recon+big` site undone, for the reason the task named as the hard part:
`cfl_ac` needs a luma row **and the row below it live at once**, and `for_rows`'s callback owns a
row's guard for exactly its own duration. `for_row_pairs` adds that one degree of freedom under the
same policy, no copy in either branch. Both borrows are immutable, so the pair cannot conflict with
itself — two subslices of one guard in the hull branch, two live guards in the per-row branch.

| `v4k_8tile` 8bpc | site (cfl_ac) | frame total |
|---|---|---|
| t=1 base | 629,080 | 7,924,706 |
| t=1 head | **67,862 (-89.2%)**, extent 10.2 B → 31,766 B | **7,363,488 (-7.1%)** |
| t=8 base/head | 629,080 / 629,080 | 22,700,725 / 22,700,725 |

### 4. Wall clock: null at every cell

9 complete rounds, 216 rows, 2 cells discarded and re-run, **zero rows under foreign load**. dav1d
1.5.4 `--framedelay 1` in the same interleaved sweep. ms/frame, median [min..max], n=9:

| cell | base | head | dav1d | Δ | disjoint |
|---|---|---|---|---|---|
| 8bpc t=1 | 319.2 [317.1..320.9] | 318.1 [315.7..320.7] | 246.5 | -0.35% | no |
| 8bpc t=2 | 198.3 [197.3..198.7] | 197.9 [197.3..199.8] | 125.1 | -0.17% | no |
| 8bpc t=4 | 107.3 [106.7..108.9] | 106.9 [106.3..108.3] | 65.6 | -0.36% | no |
| 8bpc t=8 | 68.2 [65.1..68.8] | 68.2 [65.0..69.3] | 36.3 | +0.08% | no |
| 10bpc t=1 | 359.1 [356.9..361.1] | 356.8 [355.3..360.1] | 251.2 | -0.63% | no |
| 10bpc t=2 | 206.2 [205.5..207.9] | 206.4 [204.6..207.2] | 127.6 | +0.11% | no |
| 10bpc t=4 | 111.5 [110.4..114.1] | 111.1 [110.3..113.4] | 66.8 | -0.40% | no |
| 10bpc t=8 | 72.9 [69.5..75.2] | 72.4 [69.3..73.9] | 37.6 | -0.69% | no |

**DISJOINT AT 0 OF 8 CELLS.** base/dav1d 1.295 / 1.585 / 1.635 / 1.879 (8bpc), 1.429 / 1.617 /
1.668 / 1.941 (10bpc); head within noise of each.

Paired within-round medians (the statistic the rotation exists for) run -0.69% .. +0.16%. **The
t>1 cells are a free negative control** — the census proves byte-identical extents there — and one
of them (8bpc t=4) reads p=0.039 on an exact sign test. That is this harness's false-positive rate
at n=9 over 8 tests, and it is *larger* than either t=1 cell's significance (p=0.508).

Honest bound both ways: the site-class rate predicts ~1.4 ms (0.44%) for the 561,218 registrations
removed; the observed paired median is the same order, but the spread is 4-5 ms. **"Not resolvable
at n=9", not "proven zero."**

Third independent count reduction in this campaign with no wall-clock movement (brief §6: `TinyLock`,
`block_mut` held row guards).

### 5. Gates

- Corpus `md5_inventory`: **766 PASS / 0 mismatch / 0 error / 2 skip**. Set-diff BY NAME (md5 in the
  value) vs `aarch64_md5_fixes_2026-08-07_final.tsv.zst`: 0 only-in-baseline, 0 only-in-head,
  0 differing — **CLEAN**.
- Decoded output byte-identical base vs head, 5 vectors × t{1,8}. `cargo test --release`: 0 failures
  (`decode_permutations` now 19/19 — #455 item 6 was fixed by #466).
- `mt_stress` 1/2/4/8/16 × 5 ok. `multi_decoder_pressure.sh` 12 procs × 3 iters × threads
  {1,2,4,8,16} × 5 vectors: PASS, every md5 matching the serial reference.
- disjoint-mut: soundness 25/25, `wide_exclusion` 1/1, `shard_liveness` 5/5, `narrow_release` 1/1,
  tracker unit 26/26.
- **New mechanism gate** for `for_row_pairs` (which extent it reserves is invisible in pixels), both
  branches, plus a liveness test that it yields both rows of every pair in order.
- **Teeth by mutation**, each restored from a sha256 snapshot and re-verified green: pad fill
  `last`→`0` and hull pairing row *r* with itself both fail cfl_ac parity 3/3 depths;
  `tile_threading_active()`→`false` fails the ON gate; →`true` fails the OFF child gate.
- **Pre-existing gates re-proved to still have teeth on this tree** under `--features __probe_wide`:
  the 4af62ae in-lock `state` re-read deleted, and `active()` cut to one shard — `wide_exclusion`
  FAILS on each; restored by sha256 `-c` **and** `git diff --exit-code`; green after.
- `#![forbid(unsafe_code)]` proved **actively** — planted `unsafe` block → "usage of an `unsafe`
  block"; restored, green.

### 6. Not measured

Design B (not built; only its ceiling was ever measured, by anyone). The t>1 compact copy —
`compact_read_per_row` + `BlockMut::drop`'s write-back are 2,662,876 + 1,864,816 registrations/frame
at t=8 **plus two memcpys per block**, existing only because per-row guards are needed; nobody has
priced the copy. `cfl_ac`'s scalar counterpart is unconverted (NEON dispatch only). x86_64 unbuilt
and unrun. Both 4K vectors are intra-only with LR off, so `mc_arm.rs` and `looprestoration*.rs`
register zero borrows in this grid while LR is live in 696/768 corpus vectors — the same structural
blindness the brief warns about, inherited unchanged. Neither design clears t=8 and neither can: the
untracked ceiling is 1.339 / 1.390.

Largest levers left, from this round's fresh census of ee07b00: `ctx.rs:99` 2,534,988/frame (34% of
t=1, 11% of t=8; #467 §5b, needs a concurrency argument about `f.a[t.a]`), and — new — `picture.rs:168`
at 2,336,708/frame at t=8, which is `ipred_prepare`'s LEFT-COLUMN read at w=1, i.e. 2.3 M *single-byte*
registrations. A column is not an interval, so no hull covers it; it would need a strided registration
primitive in the tracker itself.


lilith:
## Closing: its blocker is on `main` verbatim, and #482 built the form that works

Triage of the 15 open PRs, 2026-08-13. **Nothing here is lost — the structural finding, which is the
valuable half, landed.**

`docs/OWNERSHIP_MODELS.md` §1 ("`split_at_mut` on the shared picture — impossible, and the blocker is
upstream") carries this round's three reasons with the same file:line citations this PR gave:
`picture.rs:1287` (planes behind `Arc<Rav1dPictureData>`), `thread_task.rs:1237/:1416/:1436/:1451`
(shared read guards ARE the tile parallelism), `lib.rs:221` (`'static` pool threads, so no scoped
lifetime carries a split). It also carries the generalisable rule this round produced:

> **before designing around a split, check that an exclusive reference exists *at the point you
> intend to split*** … A campaign round was lost by verifying the *access pattern* was compatible and
> never checking whether the `&mut` was obtainable at all.

That sentence is about this round. It is the most-cited output of the whole Design-A exercise.

**And the "form that works" it points at was subsequently built.** #481 priced the completed
conversion, #482 wrote it soundly as `src/owned_recon.rs` — a column-compact band owned by
`Rav1dTaskContext`, handed to the kernels as `&mut [u8]`, uniqueness proved by borrowck with no
record, no lock and no `unsafe`. `OWNERSHIP_MODELS.md` §1 ends "**Do not over-generalise this
negative.** It rules out splitting *the shared plane*. It says nothing about owning a *different*
buffer — which is what actually worked."

**What is NOT on `main`:** the `for_row_pairs` helper and its `cfl_ac` census (629,080 → 67,862 at
t=1, −89.2%, frame total −7.1%; byte-identical at t=8) and its 216-row null sweep. That is a measured
negative on a helper that `main` does not have, so it is data about code that no longer exists in the
shape it was measured. It stays reachable at `pull/468/head` = **`0596f5251030337050c6a3ab853737e5e0bf7dd3`**
(`git fetch origin pull/468/head`) and is indexed in `benchmarks/pr_triage_2026-08-13.meta`.

One thing this round called correctly that is worth restating: it is the **third** count reduction in
the campaign with no wall-clock movement. `main` now knows why — cost tracks **distinct shard lines a
worker visits**, not records filed (`docs/RECT_RECORDS.md`).

PR #469, opened 2026-08-09
perf(tracker): shard picture planes by COLUMN — a w x h block is ONE borrow at every thread count (0.785x at 8bpc t=8, scaling 4.74x -> 6.13x)

lilith:
Base `main` @ ee07b00. Record: `benchmarks/strided_rect_2026-08-09.meta` + `strided_rect_gap_2026-08-09.tsv` (216 rows) + `strided_rect_bands_2026-08-09.txt` + `strided_rect_sites_2026-08-09.tsv`.

The diagnosis round on #455 named the thing that would have to change: *"a tracker registration primitive describing a strided rectangle `(offset, w, h, stride)` as ONE tracked entity — neither `h` row intervals nor one over-wide hull"*, and predicted that a change cutting only the COUNT would measure null. This builds it, and the prediction holds — the count fell 65% and the win came from the other term.

## The tie the flat block index could not break

A `w x h` block was either `h` per-row registrations (correct under tile threading, and the reason a 4K frame registers **7,924,706** borrows at t=1 but **22,700,725** at t>1) or one hull registration (cheap, but it reserves the inter-row gaps, which belong to other tile COLUMNS).

`offset >> shift` has to choose between two properties a picture plane needs at once: small blocks separate concurrent tile columns but make a strided access pay one shard line per ROW; large blocks keep the access on one line but put every tile column of a row on it. `benchmarks/tracker_blockshift_2026-08-08.meta` is the record of that being decided in both directions (shift 12 → 119.0 ms/frame at t=8, shift 14 → 72.8, shift 16 → 75.5, and going finer to separate tile columns collapsed scaling to 2.7x).

A **column** class has both properties. Every row of a rectangle occupies the same columns, so the rectangle is ONE class however tall it is; two tile columns are different classes however close their rows are.

## Mechanism

- **`RowMap`** — `offset → (column band, row group)` from one exact division by the declared row stride (Lemire's magic: a `umulh` and an `msub`, not a `udiv`). Consulted only when the instance is *both* strided and sharded, so a serial decode takes `add`'s `mask == 0` arm and pays nothing.
- **`ShardRecs::c0/.c1`** — the record's inclusive column range. Hull-overlap **and** column-overlap together are EXACT, not conservative: if the row ranges are disjoint, hull overlap needs `(rb0 - ra1)*S <= ca1 - cb0`, whose left side is ≥ S and whose right side is ≤ S−1. Anything the map cannot describe keeps `0..=COL_ANY` and degrades to today's plain hull test.
- **`StridedRows`** — the index type. Deliberately NOT a `SliceBounds` (the blanket `From` would erase the shape), and the shape travels as two scalar arguments rather than as `Bounds` fields: growing `Bounds` from 16 to 24 bytes measured **+3.6% at 8bpc t=1** on its own.
- Converted: `for_rows{,_mut}`, `narrow_guard{,_mut}`, `compact_read_per_row`, `compact_write_back_per_row`, `block_mut`, and the loop filter's `LfBlock::fill`. **`block_mut` stops copying** under tile threading — it used to read `h` rows into scratch and write them back, `2h` registrations plus two memcpys per block, purely because per-row guards were the only sound option.
- `SLOTS` 7 → 5 so the record still fits one 128-byte line. Isolated A/B of that change alone on ee07b00: 6750 vs 6760 ms — null.

## Two defects the measurement found, both now gated

1. **The wide-path avalanche.** `add_grid` escalated to the all-shards path whenever a wide record was live, so one wide record made every straddling borrow wide, which kept `state` nonzero. **188,307 wide registrations/frame at t=8** and wall **2.28x WORSE than base**. Only poison escalates now; the poison-only count is 2,220/frame and identical at t=2 and t=8, which is the check that the loop is gone.
2. **`rect_exact_for` must require `mask != 0`.** `add` dispatches `mask == 0` first, so a single-shard instance records the plain hull however well the map fits it — and telling a caller otherwise is exactly the false positive this change exists to remove. Caught by the EXISTING gate `for_rows_mut_never_reserves_an_inter_row_gap_when_tile_threading_is_on`, whose 64×4 plane is under `SHARD_MIN_LEN`.

## Gap to dav1d — n=9, 216 rows, **zero** under foreign load

dav1d 1.5.4 `--framedelay 1` in the same interleaved sweep, two-point wall fit at 2 and 20 frames, idle M4 Pro, rotating arm order.

| vector | t | base ms | head ms | dav1d ms | head/base | bands | base/ref | **head/ref** |
|---|---|---|---|---|---|---|---|---|
| v4k_8tile | 1 | 318.1 | 322.7 | 246.5 | 1.014 | OVERLAP | 1.291 | **1.309** |
| v4k_8tile | 2 | 197.8 | 166.8 | 125.2 | 0.843 | disjoint | 1.580 | **1.332** |
| v4k_8tile | 4 | 107.0 | 86.8 | 65.6 | 0.812 | disjoint | 1.631 | **1.323** |
| v4k_8tile | 8 | 67.1 | 52.7 | 35.7 | 0.785 | disjoint | 1.882 | **1.477** |
| v4k_8tile_10b | 1 | 358.3 | 363.4 | 251.3 | 1.014 | OVERLAP | 1.426 | 1.446 |
| v4k_8tile_10b | 2 | 206.1 | 188.6 | 127.6 | 0.915 | disjoint | 1.616 | 1.478 |
| v4k_8tile_10b | 4 | 110.5 | 98.6 | 66.8 | 0.892 | disjoint | 1.655 | 1.476 |
| v4k_8tile_10b | 8 | 71.2 | 60.4 | 37.2 | 0.849 | disjoint | 1.916 | 1.626 |

**Scaling t1→t8: 8bpc 4.74x → 6.13x** (dav1d 6.91x); 10bpc 5.03x → 6.01x (6.76x).

### What this does NOT do, first

**t=1 is 1.4% slower and the bands OVERLAP** — not resolvable at n=9, but the median is enough to move 8bpc t=1 from 1.291 to **1.309**, i.e. across the campaign's 1.30x bar in the wrong direction. That is the price of this change and it is not buried. The cause is priced: an UNSOUND arm with the two per-`alloc` column stores deleted reads **0.991** of head at 8bpc t=1, which would land the cell at ~1.297. Making that arm sound needs a `COLS` const parameter on `alloc`/`find` selected by the `mask == 0` dispatch, **plus** a reset of the column fields in `reprovision` — `mask` can fall back to 0 if a container shrinks below `SHARD_MIN_LEN`, and stale narrow columns would then be a MISSED overlap, not a false positive. Not attempted here.

**And the bar is still met nowhere:** 1.309 / 1.332 / 1.323 / 1.477 at 8bpc.

What it does do: at t=2 and t=4 the residual is within 0.03–0.04 of the UNTRACKED CEILING the diagnosis round measured (1.307 / 1.281 with the tracker compiled out) — essentially all of the tracker's t>1 cost is gone. #455's own arithmetic target (hold t=1, apply dav1d's ratios → 163 / 85.6 / 46.7) is met to within 2% at t=2 and t=4 (166.8, 86.8) and missed by 13% at t=8 (52.7). 10bpc moves less because a larger share of its gap was never the tracker — it starts at 1.426 at t=1 where 8bpc starts at 1.291, and that per-thread throughput difference is a separate problem this does not touch.

## Gates

- **Corpus 766/766**, set-diff BY NAME with the md5 in the value against `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`: only-in-baseline 0 / only-in-head 0 / **differing 0**.
- Full `cargo test --release`: **0 failures** (incl. `decode_permutations` 19/19). `cargo fmt --check` clean.
- Output byte-identical base vs head: one md5 per vector across t=1/2/4/8/16, both depths.
- **New MECHANISM gate**, 9 tests (`tracker_shard::rowmap_tests`) — the property is invisible in the pixels, so each asserts the mechanism, not the outcome: the map is live, a block is one class, the exact division is exact at both ends of the plane, adjacent tile columns don't conflict, a genuine overlap still does, an interval inside a row conflicts while one in the gap does not, a row-crossing interval claims every column, a wide record keeps its columns, two disjoint blocks inside ONE band don't conflict, a non-strided instance is untouched, and out-of-range shapes disable the map.
- **Teeth by mutation ×5**, each restored by sha256 + `git diff`: `find` ignoring columns → FAIL; `locate` ignoring the shape → FAIL (5 tests); wide records widening → FAIL; the division magic losing its `+1` → FAIL; a row-crossing interval claiming only column 0 → FAIL. *The first two mutations SURVIVED the gate's first draft; the two tests that catch them were added because of that.*
- Both mandatory pre-existing hazards re-planted under `--features __probe_wide`: the 4af62ae in-lock `state` re-read deleted → `wide_exclusion` FAILS 3/3; `active()` cut to one shard → FAILS 3/3 (and identically on base, so the gate is not weakened).
- `#![forbid(unsafe_code)]` proved ACTIVELY on `rav1d-safe` (planted `unsafe` → E0133 at the attribute; restored → green).
- `mt_stress` 1/2/4/8/16 × 5 ok; `multi_decoder_pressure.sh` 12 concurrent decoders × 3 iters × {1,2,4,8,16} PASS with every md5 matching the serial reference.

## Not measured

x86_64 (nothing built or run there — do not extrapolate); any vector with **loop restoration live**, so `looprestoration*.rs` registers zero borrows in this grid while LR is active in 696/768 corpus vectors; t=16 and every size below 4K in the *timed* grid; **negative strides**, where every rectangle caller falls back so the path is unexercised; `--features unchecked` (compiles, not benchmarked); Miri.


lilith:
Verification round on 4db4144 — gates in #470 (stacked on this branch). Two results, one of them blocking.

**BLOCKING — the guard hands out a `&mut` wider than it reserved.** `index_rect_mut` registers the rectangle and returns `&mut rect.hull()`. The CHANGELOG states the split on purpose (*"records the hull as the reference and the exact rectangle as the record"*) — but the *reference* is what Rust's aliasing rules bind to, so two blocks on the same rows in different tile columns are two simultaneously-live overlapping `&mut`. That is the case this PR exists to allow, and `for_rows_mut` writes through the guard (`picture.rs:1185`) while two tile workers hold overlapping hulls concurrently.

Miri: tag created over `[0x0..0xf10]`, invalidated over `[0x40..0xf50]` by the next rectangle's retag, UB on the following write. Controls in the same file **pass** — `h` per-row guards from two tile columns, and one hull guard that is also *reserved* as the hull — so it is the rectangle guard, not `DisjointMut`. `StridedRows` does not exist on ee07b00, so it arrives with the primitive.

The fix should not cost the win: the win is one *registration*, which is independent of the reference shape. A guard that keeps the base pointer and hands out `&mut` **per row** keeps the single registration with no extra atomic or lock. Five of seven call sites already index by row; `narrow_guard{,_mut}` return the guard, so the row view has to travel, and the immutable rects need it too (a `&` over the hull aliases another column's `&mut` in the gaps). I did not attempt it rather than land a half-refactor with no re-measurement.

**The predicate itself is sound, measured.** Byte-set oracle, 8 plane shapes, 1,600,000 randomized pairs: **0 missed overlaps**. 2,163 conservative false positives (0.14% of disjoint pairs), all involving a row-crossing interval — the documented `COL_ANY` fallback. It does qualify the summary line here: hull+columns is exact only when *both* records have real column ranges, not unconditionally.

**Teeth:** narrowing `cols_meet` to miss overlaps where B starts left of A leaves all **9** `rowmap_tests` green and fails the new oracle on all eight plane shapes. Restored byte-exact.

**Corpus independently re-verified: 766/766**, set-diff by name with the actual md5 as the value — 0 only-in-baseline / 0 only-in-head / 0 differing. The correctness claim here holds. I did not re-run the gap sweep or mt_stress; those numbers stand unchallenged.

lilith:
## Independent verification: the perf is real and reproduces — but six CI jobs that are green on `main` are red here, and one of them is a decode failure

Verifier round, branch [`verify/rect-audit`](https://github.com/imazen/rav1d-safe/tree/verify/rect-audit) @ `b2a743a` (record only, no source change): `benchmarks/rectverify_gap_2026-08-09.meta` + `.tsv` (168 rows, `foreign_max = 0` on every one). Idle M4 Pro, n=7, dav1d 1.5.4 in the SAME interleaved sweep, two-point wall fit. Nothing below is quoted from this PR — I re-ran it.

Branch point checked first: `git merge-base perf/strided-rect main` == `ee07b00` == `main@origin`. Not stale.

### What reproduces (all of it)

| cell | your head/base | mine | your head/ref | mine |
|---|---|---|---|---|
| 8bpc t=1 | 1.014 OVERLAP | 1.018 OVERLAP | 1.309 | **1.306** |
| 8bpc t=2 | 0.843 | 0.839 | 1.332 | **1.328** |
| 8bpc t=4 | 0.812 | 0.821 | 1.323 | **1.333** |
| 8bpc t=8 | 0.785 | 0.773 | 1.477 | **1.470** |
| 10bpc t=8 | 0.849 | 0.836 | 1.626 | 1.621 |

Scaling t1→t8 8bpc **4.83x → 6.37x** (dav1d_fd1 7.16x). Registrations/frame at t=8 **22,700,725 → 7,923,518**, and **unchanged** at t=1 — your counts to the digit, `lost=0`. Output md5 identical base-vs-head 8/8 across both vectors × t=1/2/4/8. Corpus 766/766 with a set-diff BY NAME carrying the md5 as the value: 0 only-in-baseline, 0 only-in-head, **0 differing**. `mt_stress` and the 12-decoder pressure run pass. `forbid(unsafe_code)` proved actively; both standing hazards re-planted → `wide_exclusion` FAILS 3/3 each, restored byte-exact.

Derived from my own two numbers, since the round asked for ns/registration: at t=8 the 14.78 M registrations you removed bought 14.8 ms/frame = **1.00 ns each**; at t=1 you removed **zero** and wall rose 5.6 ms = **+0.71 ns on every surviving registration**. The win is exactly where the count moved and the cost is exactly where it didn't.

One small correction: **base at 8bpc t=1 measures 1.283 here, not 1.291.** Main's best cell is already under the bar, so this lands it at 1.306 from further below than the table implies.

### CI is not mentioned anywhere in the PR body, and it is red

`main` @ ee07b00 has Clippy, Lint, Documentation, and Build & Test (macos-15-intel / ubuntu i686 / windows i686) all **green**. On this head all six **fail**. (The four `disjoint-mut CI` Test legs and `Miri (Object)` were already red on main — those are not yours.)

1. **`1usize << 32` at `tracker_shard.rs:1294` breaks every 32-bit target.** `error: this arithmetic operation will overflow` — both i686 legs. And it is wider than those two jobs show: `cargo build --release --target wasm32-wasip1` fails locally on this head and succeeds on `ee07b00`. **CI cannot see the wasm32 case** because `Cross-compile (wasm32-wasip1)` runs `cargo check`, which stops before the MIR pass that raises `arithmetic_overflow`. Third gate in this campaign that is green for a structural reason.
2. **Clippy/Lint**, `-D warnings`: `manual !RangeInclusive::contains` (1294:12) and `casting to the same type (usize -> usize)` (1316:38).
3. **Documentation**: three unresolved intra-doc links — `Self::mut_slice_as`, `Bounds::row_w`, `Bounds::rows`. The last two do not exist, and the PR body says why: the shape deliberately travels as scalar args, not as `Bounds` fields. `StridedRows`'s doc comment describes a design the PR explicitly rejected.

### The one that blocks: an intermittent `overlapping DisjointMut` in the new immutable rect path

`Build & Test (macos-15-intel, safe-simd)`, on `decode_md5_committed::lr_sgr_vectors_threaded_match_reference_md5` — 8 workers, committed LR/SGR vectors:

```
thread 'rav1d-worker-6' panicked at include/dav1d/picture.rs:568:19:
  overlapping DisjointMut:
   current:  & _[16625..32769] on ThreadId(25) at include/dav1d/picture.rs:568:19
  existing: &mut _[32768..36624]
     0: overlap_panic  at crates/rav1d-disjoint-mut/src/tracker_shard.rs:1678
    14: rav1d_filter_sbrow_deblock_cols  at src/recon.rs:3719
    15: rav1d_worker_task
```

`picture.rs:568` is the new `Rav1dPictureDataComponent::rect`. It is a **decode error**, not a wrong md5. Green on `ee07b00` (two runs) and on #468.

It **passed on re-run** of the same job, so this is a race, not an x86 break — 1 failure in 2 macos-15-intel runs. I could not reproduce it on aarch64: 1,600 runs (40 sequential + 1,560 at 24× oversubscription) plus the 12-decoder pressure sweep, all clean; no Rosetta on this box so I could not run the x86 binary. An intermittent overlap panic is the worse shape — it merges green and fails in the field.

Worth noting where the guard now comes from: `with_pixel_guard_immut`'s condition changed, but its own doc comment two lines up is untouched and says *"wide guards (`strided_slice_mut` / `narrow_guard`) are unsafe under tile threading even in the immutable case (they cover `(h-1)*stride + w` bytes and overlap with adjacent tiles' mutable ranges)"*. Narrowing the record does not narrow the reference — which is also #470's finding, arriving here as a runtime panic rather than as Miri UB.

### And the corpus gate cannot cover any of this

`examples/md5_inventory` uses `Settings::default()`, and `Settings::threads` defaults to **1**. So `tile_threading_active()` is false for all 768 vectors and every rectangle caller takes the branch `main` already took. 766/766 proves the rect *record* is pixel-inert **serially**; it says nothing about the concurrent path, which is the entire reason this change exists. The only committed test that does run threaded is the one that failed.

### What I'd want before this merges

- the `1usize << 32` fix (32-bit + wasm32), the two clippy lints, the three doc links;
- the wasm32 cross job promoted from `cargo check` to `cargo build` so it can see this class again;
- the guard narrowed to a per-row view (#470), which should also settle whether the macOS-Intel panic is a false positive or a genuine race this change newly exposes by holding a guard across `f` where `main` copied to scratch;
- `lr_sgr_vectors_threaded` looped on x86 until you have a repro rate, not a single green re-run.

The mechanism is right and the numbers are right — 0.77x at t=8 with the count down 65% is the real thing, and t=2/t=4 land within 0.03 of the untracked ceiling. None of that is in question.

lilith:
## BLOCKING — do not merge. This is UB, and it has already fired at runtime.

Converted to draft. The **mechanism is right and the perf is real** (independently reproduced below); the **guard is unsound**.

### The defect

`index_rect_mut` registers the **rectangle** but returns a `&mut [T]` over the **hull**. The CHANGELOG states this split deliberately — "records the hull as the reference and the exact rectangle as the record" — and that is precisely the bug, because **the reference is what Rust's aliasing rules bind to**, not the record. Two blocks on the same rows in different tile columns are then two simultaneously-live overlapping `&mut`, which is exactly the concurrency this change exists to enable.

Reachable on the real path: `for_rows_mut` writes through the guard at `include/dav1d/picture.rs:1185` while two tile workers hold overlapping hulls.

### It is not theoretical — it fired in CI, then passed on re-run

`Build & Test (macos-15-intel, safe-simd)` is green on `main` @ ee07b00 (two runs) and on #468, and **failed on this PR**:

```
thread 'rav1d-worker-6' panicked at include/dav1d/picture.rs:568:19:
  overlapping DisjointMut:
   current:  & _[16625..32769] on ThreadId(25)   <- the new immutable rect
  existing: &mut _[32768..36624]
   14: rav1d_filter_sbrow_deblock_cols  at src/recon.rs:3719
```

That is a **decode failure**, not a wrong md5. A re-run passed. So it is a race of unknown rate that **merges green** — the worst shape a defect can have.

Mechanism: `with_pixel_guard_immut`'s gate changed from `tile_threading_active()` to `tile_threading_active() && !pic.rect_is_exact_for::<BD>()`, so under tile threading it now holds ONE `narrow_guard` hull across the whole closure where `main` copied rows into scratch and held nothing. **Its own doc comment two lines above, untouched by the diff, already says why that is wrong:** *"wide guards (`strided_slice_mut` / `narrow_guard`) are unsafe under tile threading even in the immutable case (they cover `(h-1)*stride + w` bytes and overlap with adjacent tiles' mutable ranges)"*.

### Miri, reproduced twice independently

| case | result |
|---|---|
| CONTROL: `h` per-row guards from two tile columns (what `main` does at t>1) | pass |
| CONTROL: one hull guard, also *reserved* as the hull (what `main` does at t=1) | pass |
| two rect guards, write through both while live | **UB** |
| four tile columns held the way the decoder holds them | **UB** |

The controls passing is load-bearing: this is the **rectangle guard**, not `DisjointMut` generally.

### The perf is real and worth saving

Independently re-measured, n=7, dav1d in the same interleaved sweep, idle box — #469's own numbers land within 0.01 at every cell:

| t | base | head | head/dav1d | bands |
|---|---|---|---|---|
| 1 | 316.2 | 321.8 | 1.306 | OVERLAP |
| 2 | 198.2 | 166.3 | **1.328** | disjoint |
| 4 | 106.5 | 87.4 | **1.333** | disjoint |
| 8 | 65.4 | 50.6 | **1.470** | disjoint |

Scaling 4.83x → **6.37x** (dav1d 7.16x). Registrations at t=8: 22,700,725 → 7,923,518 (**−65%**).

**And it confirms the campaign's "count is not cost" rule in both directions inside one change:** at t=8 the 14.78 M removed registrations bought 14.8 ms = 1.00 ns each; at t=1 **zero** registrations were removed and wall rose 5.6 ms = **+0.71 ns on every surviving registration** (the row-map division plus two column stores, paid by a path that removes nothing). The win is exactly where the count moved; the regression is exactly where it did not.

### The fix, which should preserve the win

A **per-row view on the guard** — keep the ONE registration, hand out per-row references. No extra atomic, no extra lock, so the measured win should survive.

Scope: 7 call sites. Five index by row inside the function and are mechanical (`for_rows`, `for_rows_mut`, `compact_read_per_row`, `compact_write_back_per_row`, `LfBlock::fill`). `narrow_guard`/`narrow_guard_mut` return the guard, so the row view has to travel to their callers. **The immutable rects need it too** — a `&` over the hull aliases another column's `&mut` in the gaps.

Also fix before merge: the PR body's headline says the test is "EXACT, not conservative". `ShardRecs::c0`'s own doc is correct and the body is not — hull+columns is exact only when **both** records carry real column ranges. An interval running off the end of its row registers `0..=COL_ANY` and degrades to the plain hull test: conservative, never missing, but it produced 2,163 false positives in 1.6 M oracle pairs (0.14%). Spurious `overlapping DisjointMut` panics are **decode failures**, and #467 measured 8-9 in 24 concurrent runs from that shape.

### Gates now exist for this

#470 (stacked on this branch) adds the brute-force byte-set oracle and the Miri aliasing gate — neither existed before (`grep -c 'proptest|oracle|brute'` over the crate returned 0). Oracle soak: 1.6 M randomized pairs across 8 plane shapes, **0 missed overlaps**, and a planted `cols_meet` mutation that **9/9 shipped `rowmap_tests` miss** is caught on 8/8 shapes.

lilith:
## The guard is fixed in #472 — and the macOS-Intel panic is NOT the aliasing bug

Row-view fix for the blocking finding: **#472** (`fix/rect-rowview`, based on this branch).
`DisjointMutRect` / `DisjointImmutRect` keep the ONE registration and hand out one **row** at a
time; no reference is created at construction, so every reference is a subset of what the
tracker reserved. Miri: all nine cases pass, including the two that were UB. Corpus 766/766
set-diff BY NAME, 0 differing. It also clears the `1usize << 32` 32-bit hard error, the two
clippy findings and the three doc links, so Lint and Documentation are green there.

**Two of the seven call sites could not be converted, and that is the price.**
`narrow_guard{,_mut}` hand the caller a FLAT slice over the hull plus a stride — that shape *is*
a hull reference, and narrowing the record does not narrow it. They revert to a plain hull
borrow and their callers (`with_pixel_guard_immut`, `block_mut`) revert to gating on
`tile_threading_active()`. `block_mut` therefore loses this PR's copy elision; the rectangle
reaches it one level down instead, so a block costs 2 registrations rather than `2h`.

### The CI failure is a GENUINE overlap, and the site is x86-only

A fresh failure on the merge commit c1355c4 (run 31294761102, job 93197936544) carries a full
stack that the verification round did not have:

```
 current:    & _[241..16385] on ThreadId(22)
existing: &mut _[16384..20240]
  3: add_wide<false>          tracker_shard.rs:2286
  4: add_rect_immut           tracker_shard.rs:1709
  8: compact_read_per_row     picture.rs:1254
  9: loopfilter_sb_dispatch   src/safe_simd/loopfilter.rs:5134
 14: rav1d_filter_sbrow_deblock_cols
```

Two facts follow **without knowing the plane stride**:

1. The last byte of a strided rectangle's hull is by construction the last element of its last
   ROW, never a gap. A's hull ends at 16385, so A covers byte 16384; B covers `[16384, 20240)`.
   **Their covered byte sets intersect at 16384.** It is neither a `COL_ANY` false positive nor
   an artefact of the hull reference — so **#472 does not remove it**.
2. `loopfilter_sb_dispatch` in `src/safe_simd/loopfilter.rs` is `#[cfg(target_arch = "x86_64")]`
   (wasm32 shares it); aarch64 goes to `loopfilter_arm`, which never calls
   `compact_read_per_row`. **That is why 1,600 aarch64 runs could not reproduce it — the failing
   call site is not compiled on that box.**

What this PR changed at that site is the WINDOW, not the extent. `main` reads the same tap
window through `h` per-row guards, each live only for its own memcpy; the rectangle reserves all
`h` rows for the whole copy loop. The bytes are identical, so `main` has the same unsynchronized
overlap and only detects it when the conflict lands inside the one row whose guard is live at
that instant. **This is a property of "one registration for a multi-row access", not of the hull
reference.** The overlapping column is the loop filter's deliberate over-read: 8bpc H-luma reads
`7 + 9 = 16` columns because the kernel loads 16 B/row, while the tap ladder only reaches
p6..q6 — columns +7 and +8 are read and never used, and the same file already documents
narrowing this window once for exactly this reason.

### One correction to the summary line

"EXACT, not conservative" holds only between two records that BOTH carry a real column range —
`ShardRecs::c0`'s own doc was right and the body was not. Re-measured at
`RECT_ORACLE_ITERS=200000`: 1,600,000 pairs, **0 missed overlaps**, 2,163 conservative false
positives (0.14% of disjoint pairs), every one involving a row-crossing interval. #472 rewrites
that doc, records why the number is a budget (a false positive is a decode failure), and writes
down the exact fix it deliberately did NOT take.

lilith:
## Correction: my root-cause attribution above was WRONG

I attributed the `macos-15-intel` panic to the hull-reference aliasing defect. **It is not that bug**, and #472's investigation proved it. Correcting the record before anyone acts on my earlier comment.

The fresh failure on the merge commit c1355c4 (run 31294761102, job 93197936544) has the full stack:

```
current:  & _[241..16385]        <- compact_read_per_row -> add_rect_immut
existing: &mut _[16384..20240]
```

Two things follow without needing the plane stride:

1. **It is a genuine overlap**, not a `COL_ANY` false positive and not an artefact of the hull reference. The last byte of a `StridedRows` hull is by construction the last element of its last *row*, never a gap — so A really covers byte 16384, and B covers `[16384, 20240)`.
2. **The failing call site is `src/safe_simd/loopfilter.rs::loopfilter_sb_dispatch`, which is `#[cfg(target_arch = "x86_64")]`.** aarch64 routes to `loopfilter_arm`, which never calls `compact_read_per_row`.

That second point explains the thing that should have made me suspicious: the verifier's **1,600 aarch64 runs** (40 sequential + 1,560 at 24x oversubscription) plus a 12-decoder pressure sweep could not reproduce it. Not because it is rare on this box — because **the failing code is not compiled on this box.** I read "could not reproduce on aarch64" as "rare race" when it meant "wrong architecture".

**The real mechanism:** #469 changed the *window*, not the extent. `main` reads the same tap window through `h` per-row guards, each live for exactly one memcpy; the rectangle reserves all `h` rows for the whole copy loop. That is a property of "one registration for a multi-row access", so **the row-view fix in #472 does not remove it** — as #472 confirms.

The overlapping column is the loop filter's deliberate over-read: 8bpc H-luma reads 7+9 = 16 columns because the kernel loads 16 B/row while the tap ladder ends at q6, so +7 and +8 are read and unused. Fixing that is x86-only work and could not be run here (no Rosetta on this box).

Repro rate, honestly: **n=3 total** — 1 of 2 `macos-15-intel` runs on 4db4144, 1 of 1 on c1355c4, and 0 of 1,600 aarch64 runs at a site not compiled there.

### What still stands from my earlier comment

The hull-reference aliasing defect **is real and is separate** — Miri flags `writing_through_both_guards_while_both_are_live` and `four_tile_columns_the_way_the_decoder_holds_them` as UB while both controls pass. #472 fixes that one. This PR now has *two* independent blockers, and only one of them has a fix in flight.

### Also from #472, worth recording here

- **#469's own t=1 fix does not reproduce.** The `COLS` const-parameter arm that #469 credits with "0.991 with the two column stores deleted" was built exactly as described and measured at n=9 across three cells: 0.9966 / 1.0139 / 1.0037, **every band overlapping, sign inconsistent.** Reverted.
- **c1355c4 does not build on 32-bit at all**: `1usize << 32` at `tracker_shard.rs:1294` is an `arithmetic_overflow` hard error on both `i686-unknown-linux-gnu` and `wasm32-wasip1`. Fixed in #472 by computing in `u64`.

lilith:
## Closing: the GOAL was reached twice, by two different mechanisms — neither of them this one

Triage of the 15 open PRs, 2026-08-13. **And a correction to how this PR has been described**, which
is worth more than the closure.

### The supersession claim I was handed was imprecise, and I checked it

The triage brief said "#469's column-sharding is the shape #482's column-compact `ReconBand` landed."
That is not right. This PR's mechanism is a **tracker** change — `RowMap`, per-record `c0`/`c1` column
bands, `cols_meet`, `COL_ANY`, `StridedRows`. **None of those symbols exist anywhere on `main`**
(`git grep` over the whole tree returns only two prose mentions, in `bounds_map_2026-08-10.meta` and
`OWNERSHIP_MODELS.md`). Column sharding was never landed and is not what shipped.

What actually reached this PR's stated goal — *a `w x h` block is ONE borrow* — is two later changes,
covering two different populations:

| population | what landed | mechanism |
|---|---|---|
| reconstruction | **#482** `src/owned_recon.rs` | an **owned** column-compact band on `Rav1dTaskContext`, `&mut [u8]`, uniqueness by borrowck — **no record at all**, one band per WORKER not per tile |
| filter chain + CDEF | **#505**, default since **#513** (2026-08-11) | an **exact strided-rectangle record**, `add_rect_{mut,immut}`, scan cost unchanged |

#482 removes the registrations; it does not make them column-aware. The change that generalises
"one record for a rectangle" is #505/#513, and `Cargo.toml` now declares `__lf_rect` and `__rows_rect`
as retained no-ops saying "the rectangle is now the DEFAULT path (2026-08-11)".

### The other reason this cannot merge

It shipped **Miri UB**: an exact record behind a **hull reference**, i.e. two tile columns' rectangles
handing out two live overlapping `&mut`. #470 found it and #472 fixed the shape — which is exactly the
`DisjointMutRectGuard::row_mut(&mut self)` API `main` has today at
`crates/rav1d-disjoint-mut/src/lib.rs:1615`. So the *reference* half of this design also landed, via
#472 → #505.

### What IS on `main` from this round

* the **wide-path avalanche** (2.28x slower than base when `add_grid` escalated on any live wide
  record) — `OWNERSHIP_MODELS.md:68`, `bounds_map_2026-08-10.meta:188` and `:214`;
* the verdict and its scope — `BOUNDS_MAP.md:385` has a section per this PR, and `:222` explicitly
  says *"The '#469 rectangle → BLOCKED' verdict is about the HULL, and #505 has…"*, so nobody carries
  "the rectangle is blocked" forward.

### What is not

`strided_rect_*_2026-08-09.{meta,tsv,txt}` — the 216-row sweep, the bands file and the site census.
They measure a tracker that does not exist on `main`, so landing them would put a record in
`benchmarks/` that no current code can reproduce. Reachable at `pull/469/head` =
**`c1355c423c6b3844d006f92e6dd199fe2393dcd6`**; indexed in `benchmarks/pr_triage_2026-08-13.meta`.

PR #470, opened 2026-08-09
test(disjoint-mut): oracle + Miri gates for the rect primitive — predicate sound over 1.6M pairs, but the guard hands out an aliasing &mut

lilith:
Stacked on #469 (base `perf/strided-rect` @ 4db4144). Adds the two gates that PR's `Not measured` list names — a randomized oracle and Miri — and one of them fails.

## BLOCKING: the rectangle guard hands out a `&mut` wider than it reserved

`index_rect_mut` registers the **rectangle** and returns a `&mut [T]` over `rect.hull()`. The CHANGELOG entry states the split deliberately — *"records the hull as the reference and the exact rectangle as the record"* — and that is the defect: the reference is what Rust's aliasing rules apply to, so two blocks on the same rows in different tile columns are two **simultaneously live overlapping `&mut`**. That is the case the row map exists to allow, so it is not a corner.

```
A = rect(start 0,  w 16, h 16, stride 256)  -> hull [0x000 .. 0xf10)
B = rect(start 64, w 16, h 16, stride 256)  -> hull [0x040 .. 0xf50)
bytes: columns 0..=15 vs 64..=79 of the same 16 rows — disjoint
refs:  the two hulls overlap by 3,792 bytes
```

Miri, `crates/rav1d-disjoint-mut/tests/rect_hull_aliasing.rs`:

```
error: Undefined Behavior: trying to retag from <131284> for Unique permission
       at alloc42031[0x40], but that tag does not exist in the borrow stack
help: <131284> was created by a Unique retag at offsets [0x0..0xf10]
help: <131284> was later invalidated at offsets [0x40..0xf50] by a Unique retag
```

| under Miri | |
|---|---|
| CONTROL — `h` per-row guards from two tile columns, 32 live (what `main` does at t>1) | **pass** |
| CONTROL — one hull guard, which the tracker also RESERVES as the hull (what `main` does at t=1) | **pass** |
| two rect guards created, earlier one never used again | pass |
| two rect guards, write through both while both live | **UB** |
| four tile columns held the way the decoder holds them | **UB** |

The controls are the point: this is the rectangle guard, not `DisjointMut` generally. `StridedRows` does not exist on `main` (`git show ee07b00:…/lib.rs | grep -c StridedRows` → 0), so the hazard arrives with the primitive.

**Reachable on the real path**, not just in the test: `for_rows_mut` does `f(row, &mut guard[idx..][..w])` (`include/dav1d/picture.rs:1185`) — it writes through the hull guard while holding it, and two tile workers do that concurrently on overlapping hulls by design.

### The fix does not cost the win
The win is **one registration** per block; the reference shape is independent of it. A guard that keeps the base pointer and hands out `&mut` **per row** keeps the single registration and makes the references disjoint (rows of one rect are disjoint from each other and, by the tracker, from every other borrow). No extra atomic, no extra lock.

Scope: 7 call sites. Five index by row inside the function and are mechanical (`for_rows`, `for_rows_mut`, `compact_read_per_row`, `compact_write_back_per_row`, `LfBlock::fill`). Two — `narrow_guard` / `narrow_guard_mut` — return the guard to their callers, so the row view has to travel. Immutable rects need it too: a `&` over the hull aliases another column's `&mut` in the gaps.

I did not attempt it here rather than land a half-refactor with no re-measurement.

## The predicate itself is sound — measured, not argued

`crates/rav1d-disjoint-mut/tests/rect_oracle.rs`. The oracle is a literal byte set: paint every byte A covers, ask whether any byte B covers is painted. Nothing about hulls, columns, bands or shards is transcribed from the implementation. Eight plane shapes (row-map floor 16, non-powers-of-two 100/257, a 4K luma row and its 10-bit twin), randomized rect-vs-rect and rect-vs-interval, `RECT_ORACLE_ITERS=200000`:

**1,600,000 pairs — 63,020 overlapping / 1,536,980 disjoint — 0 missed overlaps.**

2,163 conservative false positives (0.14% of disjoint pairs), **all** involving a row-crossing interval. That is the documented `COL_ANY` fallback to the plain hull test, but it does qualify this PR's summary line, which says the pair is *"EXACT, not conservative"* without the caveat that this holds only when **both** records have real column ranges. The gate splits on exactly that line: missed overlaps are a hard zero, false positives are permitted only where at least one shape crosses a row, so a false positive between two describable shapes fails.

Worth noting because a spurious `overlapping DisjointMut` panic is a decode failure, and it is the hazard #467 hit (8–9 in 24 concurrent runs).

## Teeth

Narrow `cols_meet` from `a0 <= b1 && b0 <= a1` to `a0 <= b0 && b0 <= a1`, so it misses every overlap where B starts left of A — a missed overlap **is** two aliasing `&mut`:

- all **9** shipped `rowmap_tests`: **green**
- `rect_oracle`: **fails on all eight plane shapes**, labelled `MISSED OVERLAP (UNSOUND)`

Restored byte-exact (sha256 `06c5d452…` before and after, `git diff` clean, suite green). The shipped mechanism suite had a hole in the soundness direction; this closes it.

## Also verified
- Corpus **766/766**, set-diff BY NAME with the actual md5 as the value vs `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`: 0 only-in-baseline / 0 only-in-head / 0 differing. #469's correctness claim holds.
- `cargo test -p rav1d-disjoint-mut` green (35 + 25 + 6 + 5 + 2 + 1 + 1); `cargo fmt --check` clean.

## Not done
I did **not** re-run the dav1d gap sweep, `mt_stress`, or `multi_decoder_pressure.sh` — #469's own numbers stand unchallenged and re-running them adds nothing while the reference-aliasing question is open. No x86_64. No fix. `rect_hull_aliasing` passes on stable by construction (UB is not a runtime failure) — it is a Miri gate and CI must run it under Miri to have any teeth.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,26 @@ All notable changes to the `rav1d-safe` crate are documented in this file. Forma
 
 ## [Unreleased]
 
+### Added
+- **Two soundness gates for the `StridedRows` rectangle borrow**
+  (`crates/rav1d-disjoint-mut/tests/rect_oracle.rs`,
+  `crates/rav1d-disjoint-mut/tests/rect_hull_aliasing.rs`).
+  - `rect_oracle` checks the overlap predicate against a brute-force BYTE-SET
+    oracle over randomized geometries on eight plane shapes (row-map floor 16,
+    non-powers-of-two, a 4K luma row and its 10-bit twin). 1,600,000 pairs:
+    **0 missed overlaps**, and the 2,163 conservative false positives (0.14% of
+    disjoint pairs) all involve a row-crossing interval — the documented
+    `COL_ANY` degradation. Between two shapes the row map CAN describe, the
+    predicate is exact, and the test fails if that stops being true. Mutation
+    proof: narrowing `cols_meet` to test only one edge (a MISSED-overlap bug,
+    i.e. two aliasing `&mut`) leaves **all 9 shipped `rowmap_tests` green** and
+    fails this gate on all eight plane shapes.
+  - `rect_hull_aliasing` records, under Miri, that `index_rect_mut` reserves the
+    rectangle but hands out a `&mut` over the HULL, so two blocks on the same
+    rows in different tile columns are two live overlapping `&mut`. Controls in
+    the same file (per-row guards; one hull guard that is also reserved as the
+    hull) pass, isolating it to the rectangle guard. See the file's module doc.
+
 ### Changed
 - **The borrow tracker shards a picture plane by COLUMN, and a `w x h` block is
   ONE tracked borrow at every thread count**
--- a/crates/rav1d-disjoint-mut/tests/rect_hull_aliasing.rs
+++ b/crates/rav1d-disjoint-mut/tests/rect_hull_aliasing.rs
@@ -1,0 +1,153 @@
+//! Does a rectangle guard hand out MORE than the tracker reserved?
+//!
+//! `index_rect_mut` registers the rectangle (so the inter-row gaps stay
+//! available to other tile columns — the whole point of `StridedRows`) but the
+//! reference it returns is `rect.hull()`, i.e. the gaps included. Two blocks on
+//! the same rows at different columns are therefore accepted by the tracker
+//! while their `&mut [u8]` ranges overlap.
+//!
+//! Run under Miri to decide it:
+//! `cargo +nightly miri test -p rav1d-disjoint-mut --test rect_hull_aliasing`
+
+use rav1d_disjoint_mut::{DisjointMut, StridedRows, set_parallelism};
+
+/// Small enough for Miri, big enough to be sharded (`SHARD_MIN_LEN` = 64 KiB)
+/// and to get a row map (stride >= 16).
+const STRIDE: usize = 256;
+const ROWS: usize = 256;
+const LEN: usize = STRIDE * ROWS;
+
+fn rect(start: usize, w: usize, h: usize) -> StridedRows {
+    StridedRows {
+        start,
+        w,
+        h,
+        stride: STRIDE,
+    }
+}
+
+/// First, without Miri: state the geometry as a plain fact.
+///
+/// This asserts only arithmetic — it passes on any toolchain and documents
+/// exactly what the Miri test below is exercising.
+#[test]
+fn the_hulls_of_two_accepted_rectangles_overlap() {
+    let a = rect(0, 16, 16);
+    let b = rect(64, 16, 16);
+    let a_hull = a.start..a.start + (a.h - 1) * a.stride + a.w;
+    let b_hull = b.start..b.start + (b.h - 1) * b.stride + b.w;
+    assert!(
+        a_hull.start < b_hull.end && b_hull.start < a_hull.end,
+        "the two hulls must overlap for this to be a question: {a_hull:?} vs {b_hull:?}"
+    );
+    // ... while the BYTES each one actually covers are disjoint (columns
+    // 0..=15 versus 64..=79 of the same sixteen rows).
+    for row in 0..16 {
+        let ar = a.start + row * STRIDE..a.start + row * STRIDE + a.w;
+        let br = b.start + row * STRIDE..b.start + row * STRIDE + b.w;
+        assert!(ar.end <= br.start, "row {row}: {ar:?} vs {br:?}");
+    }
+}
+
+/// The tracker accepts both — that part is CORRECT and is what the row map is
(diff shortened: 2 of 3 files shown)


lilith:
## Second verifier: both gates reproduce exactly, and the defect has now also fired at runtime in CI

Re-ran everything here independently on an M4 Pro (record: [`verify/rect-audit`](https://github.com/imazen/rav1d-safe/tree/verify/rect-audit) @ `b2a743a`, `benchmarks/rectverify_gap_2026-08-09.meta`).

**Miri matrix — identical, run one test at a time since Miri aborts on the first UB:**

| case | mine |
|---|---|
| `the_hulls_of_two_accepted_rectangles_overlap` | pass |
| `the_tracker_accepts_both` | pass |
| CONTROL per-row guards, two tile columns | pass |
| CONTROL one hull guard | pass |
| `writing_through_both_guards_while_both_are_live` | **UB** |
| `four_tile_columns_the_way_the_decoder_holds_them` | **UB** |

`trying to retag from <N> for Unique permission … that tag does not exist in the borrow stack`, raised at `DisjointMutGuard::deref_mut` (`lib.rs:322`). Controls passing is the load-bearing part and it holds. `StridedRows` / `index_rect_mut` are 0 occurrences on `ee07b00`, so attribution to the new primitive is clean.

**Oracle — to the digit.** `RECT_ORACLE_ITERS=200000`: 63,020 overlapping / 1,536,980 disjoint, **0 missed overlaps**, 2,163 false positives (0.14%), all row-crossing, 1.75 s. `tracker_shard.rs` sha256 `06c5d452…d23539` matches yours.

**Teeth, two mutations, each restored byte-exact (`git diff --exit-code` clean):**

| mutation | shipped `rowmap_tests` | `rect_oracle` |
|---|---|---|
| M1 — yours, `cols_meet` → `a0 <= b0 && b0 <= a1` | **9/9 GREEN** | FAIL, `MISSED OVERLAP (UNSOUND)` on 8/8 shapes |
| M2 — mine, `StridedRows::shape` claims `w - 1` columns | 8/9 (`adjacent_tile_columns_on_the_same_rows_do_not_conflict` FAILS) | FAIL, `MISSED OVERLAP (UNSOUND)` |

M1 confirms your central claim. M2 is worth recording alongside it: the shipped tests are not *uniformly* toothless — a width mutation is caught, a predicate mutation is not. The distinction is the useful one: hand-picked geometries gate the *shape* the author thought about, and only randomized byte-set comparison gates the *predicate*.

### One coverage gap in `rect_oracle`, and CI just walked into it

`conflicts()` and `hold()` both take **mutable** borrows only. `index_rect` / `rect_as` — the IMMUTABLE rectangle path — is never exercised, and it uses a different scan mask (`find::<false>` scans `occupied & self.mutable`, not `occupied`). The oracle is also single-threaded by construction.

That matters because the defect has now shown up at runtime on the path the oracle does not cover. `Build & Test (macos-15-intel, safe-simd)` on #469's head failed `decode_md5_committed::lr_sgr_vectors_threaded_match_reference_md5` (8 workers, committed LR/SGR vectors) with

```
thread 'rav1d-worker-6' panicked at include/dav1d/picture.rs:568:19:
  overlapping DisjointMut:
   current:  & _[16625..32769]        <- the new immutable rect
  existing: &mut _[32768..36624]
```

— a **decode error**, green on `ee07b00` and on #468, and it passed on re-run, so it is a race with an unknown rate rather than a deterministic x86 break. Adding an immutable arm (`hold` a mutable interval, try an immutable rect and vice versa) would put that pair under the oracle. Full detail in my comment on #469.

### Two things I'd add to your list

- **The Miri job exists and is already red on `main`.** `Miri (Object)` is a pre-existing failure on `ee07b00`, so `rect_hull_aliasing` currently lands in a job nobody can read a signal from. Your trap #4 ("if CI does not run Miri, that file is decoration") is right but understated — it *does* run, and it is decoration for a different reason. Fixing that job, or adding a small `rav1d-disjoint-mut`-only Miri leg, is the cheap unlock.
- **The corpus gate cannot see this class at all.** `md5_inventory` uses `Settings::default()` and `threads` defaults to 1, so all 768 vectors decode with `tile_threading_active() == false` and every rectangle caller takes the branch `main` already took. Your "corpus claim holds" is correct and I reproduce it (0 differing on a name-keyed md5 set-diff) — it just does not bear on the concurrent path.

Your call not to land a half-refactor was the right one. Landing the per-row view without re-measuring would have put an unmeasured change under a measured headline, and the t=1 cell is already the fragile one (bands OVERLAP at n=7 in my sweep too).

PR #471, opened 2026-08-09
perf(tracker): bounded spin then yield — the shipping build spun without bound

lilith:
User call, and it is right: **the spinlock is bad, and the shipping build is worse than "no backoff."**

`TinyLock::lock_slow` (`crates/rav1d-disjoint-mut/src/tracker_shard.rs:471`) spins forever on `spin_loop()`. The `std::thread::yield_now()` existed **only** behind `__probe_lock_backoff`, which is not in `default` — so no shipping build has ever yielded.

## Why this is not the "TinyLock backoff measured null ×2" result again

Backoff and yielding-to-the-scheduler are different mechanisms, and the prior null tested backoff *as a probe arm on a build that was already spinning unbounded in production*. What is bad here is not the cost of spinning; it is what an unbounded spinner does when the holder is not running:

- The critical section is a few instructions and contention is **~0.02%** (~23,009 slow paths per 6-frame run against 22,700,725 registrations). The spin almost never runs — so when it does, it is because something abnormal happened to the holder.
- **Apple silicon is heterogeneous** (M4: 8 P-cores + 4 E-cores). A spinner on a P-core waiting for a holder the scheduler placed on an E-core burns a fast core for a multiple of the hold time.
- A **preempted** holder makes an unbounded spinner burn its whole quantum — ~1 ms against a hold time of nanoseconds.

## The evidence is already in our own sweeps

The signature of this pathology is a fat tail with a normal median, and the t=8 cells show exactly that: 8bpc t=8 bands of **[67.8..114.6]** and **[73.2..112.8]** around medians near 70. Steady-state cost does not produce a 45 ms tail; occasional catastrophic spin episodes do.

**So judge this on p90/max and band width, not on the median.** A median-only A/B will read as null and would be the wrong conclusion.

## The fast path is untouched by construction

`lock`, `try_lock` and `unlock` remain a single atomic each. `try_lock`'s freedom from any branch-and-link is load-bearing for the frameless leaf `add` — the comment at `:456` records that a `bl` there forces a 0x80-byte frame plus three `stp`/`ldp` pairs on **every** registration. The yield lives only inside the already `#[cold] #[inline(never)]` slow path, so t=1 cannot regress from it.

`no_std` keeps the old unbounded spin (there is nowhere to yield to); `std` is this crate's default feature, so this is not a no_std tradeoff.

Also deletes `__probe_lock_backoff` / `probe-lock-backoff`: the probe gated exactly the behaviour that is now default, so keeping it would leave a no-op arm for a future session to burn a round on.

## Verified

- Corpus **766/766** — `decode_md5_verify` 14/14 in 34.2 s.
- `rav1d-disjoint-mut` suite green: 26 lib + 25 soundness + 5 shard_liveness + `wide_exclusion` **under `--features __probe_wide`** (without that feature its anti-vacuity assertion is compiled out) + `narrow_release`.
- Builds clean with and without default features.

## Owed

**Timing is NOT measured.** Another campaign holds the box, and a contended-lock change measured under foreign load would be exactly the wrong thing to trust. The A/B is owed and must report **p90 and max per cell, with band widths**, at t=4/8/16 where contention is highest — not medians alone. Deliberately not merged until then.


lilith:
## Measured. **The premise is refuted; a smaller, real win survives at t=16 only.**

Timing is no longer owed. Record on the branch @ `5807083`:
`benchmarks/park_not_spin_2026-08-09.meta` + `_gap_` (310 rows) + `_bands_` +
`_lockstats_`. Idle M4 Pro (8P+4E), macOS 26.5.2, rustc 1.97.1, base `main` @ `ee07b00`,
n=16 rounds, arms interleaved with rotating order, two-point wall fit, STRICT idle gate —
**every committed row has `foreign_max = 0`** (the box was shared with two other campaigns;
the gate discarded and re-ran affected cells rather than admitting them). No `nice` on any
timed run. **Not merged.**

### What this does NOT do — the criterion the PR asked for reads null

The PR said: *"judge this on p90/max and band width, not on the median."* Judged that way at
t=4 and t=8, **this change measures null.** It survives on a result it did not predict.

| vec | t | n | base med | head med | r_med | r_p90 | r_max | band base | band head |
|---|---|---|---|---|---|---|---|---|---|
| v4k_8tile | 1 | 16 | 318.39 | 319.69 | 1.004 | 1.000 | 0.998 | 25.50 | 19.00 |
| v4k_8tile | 2 | 16 | 198.25 | 198.42 | 1.001 | 1.005 | 0.999 | 3.61 | 3.00 |
| v4k_8tile | 4 | 16 | 107.39 | 106.97 | 0.996 | 0.994 | 0.979 | 5.44 | 2.78 |
| v4k_8tile | 8 | 16 | 67.44 | 67.86 | 1.006 | 0.998 | 1.003 | 5.17 | 5.94 |
| **v4k_8tile** | **16** | 16 | **72.97** | **69.69** | **0.955** | **0.944** | **0.944** | 6.06 | 4.94 |
| v4k_8tile_10b | 1 | 16 | 359.94 | 360.00 | 1.000 | 1.001 | 1.000 | 4.06 | 4.28 |
| v4k_8tile_10b | 2 | 16 | 206.61 | 206.53 | 1.000 | 0.999 | 1.000 | 4.89 | 4.89 |
| v4k_8tile_10b | 4 | 15 | 110.89 | 112.00 | 1.010 | 0.989 | 1.006 | 6.06 | 5.78 |
| v4k_8tile_10b | 8 | 14 | 71.19 | 72.58 | 1.020 | 1.018 | 1.055 | 4.56 | 8.78 |
| **v4k_8tile_10b** | **16** | 14 | **77.19** | **74.14** | **0.960** | **0.969** | **0.971** | 4.39 | 6.50 |

(10bpc t=8/t=16 are n=14 rather than 16 — the box went busy indefinitely. Stated, not padded.)

**There is no fat tail to collapse.** Base band widths at t=8 are **7.7%** of the median at
8bpc and **6.4%** at 10bpc — not the ~60% the rationale assumed. Head's bands are not
systematically narrower; at 10bpc t=8 head's band is nearly **twice as wide** (12.1% vs 6.4%),
and at 8bpc t=8 it is wider too.

**The cited band evidence has no source and does not reproduce.** `[67.8..114.6]` and
`[73.2..112.8]` appear in **no committed benchmark TSV** in this repo (I recomputed per-cell
bands over every gap-shaped TSV in `benchmarks/`) and in **no issue or PR body except this
one**. The fattest committed t=8 bands are in `scaling_gap_dav1d_2026-08-08.tsv` — `plain`
`[72.33..97.89]` (33%) — a file whose own meta already tags it `ALLOW_LOAD=1` with foreign
qemu above the gate in 36 of 60 rows. **In that same file dav1d, which contains none of our
locking, bands `[34.28..57.78]` (64%) and `[32.11..43.06]` (32%).** Those fat t=8 bands are a
property of a loaded box, not of `TinyLock`.

### What it DOES do

**t=16 — the only cell on this box where threads exceed cores — moves 0.955 / 0.960 on the
median**, and p90 and max move by the same factor: the whole distribution shifts, nothing is
being clipped. 8bpc bands are base `[69.22..75.28]` vs head `[66.11..71.06]`, touching over
1.8 ms of a 6 ms span. t=1/2/4/8 are null on every statistic, both depths. That is the
mechanism stated as a measurement rather than an argument.

### The contention figure in the PR body is the wrong counter

"~0.02%, 23,009 slow paths" is `wide_probe::N_SLOW`, which counts
`BorrowTracker::add_slow` (poisoned / live-wide-record / multi-block). That is a **different
event from lock contention**. New throwaway probe `--features probe-lockstats`
(`rav1d_disjoint_mut::lock_probe`: one `fetch_add` per *contended* acquisition, none on the
fast path, none inside the spin loop) measures the real thing — and it barely perturbs
(71.9 ms/frame at 8bpc t=8 against 67.4 un-probed, unlike `probe-count`, which runs the same
cell at 6,537 ms/frame and whose own 19.7% contention number is entirely self-inflicted).

v4k_8tile 8bpc, 6 frames, measured here: **22,700,725** registrations/frame, **45,401,450**
lock acquisitions/frame, and 150,858–198,316 contended acquisitions per 6-frame run
→ ~27,800/frame = **0.122% of registrations / 0.061% of acquisitions — six times the quoted
rate.**

One `spin_loop()` is **8.28 ns net** on this host (`isb`, not a nop; 20 M-iteration loop, best
of 5, minus an identical empty loop). That prices the counters:

| arm | t | contended | spin iters | yields | max spins in one acquisition |
|---|---|---|---|---|---|
| base | 4 | 191,445 | 1,688,666 | 0 | 1,083 (~9.0 µs) |
| base | 8 | 150,858 | 1,322,714 | 0 | 1,387 (~11.5 µs) |
| base | 16 | 192,249 | 2,033,586 | 0 | **13,262 (~110 µs)** |
| head | 4 | 167,968 | 1,297,033 | 3,044 | 900 |
| head | 8 | 198,316 | 1,528,031 | 2,657 | 841 |
| head | 16 | 270,246 | 2,175,310 | 3,638 | 1,179 |

- **Liveness: the 64-spin bound really is reached** — 2,657–3,638 yields per 6-frame run,
  ~1.4% of contended acquisitions. The change is not inert. The other **98.6% resolve inside
  64 spins**, which is exactly why t≤8 cannot move.
- **The unbounded spin is real but small.** Worst single acquisition ~11.5 µs at t=8 and
  ~110 µs at t=16 — two orders of magnitude short of a preemption quantum. The predicted
  45 ms tail was never going to be there.
- Aggregate spin work at t=16 falls ~23% (2.21 M → 1.74 M iterations per 6 frames) = 0.64
  ms/frame of CPU, against a **3.3 ms/frame wall win** — so most of the t=16 gain is the
  descheduled holder getting a core back, not the waiter's own burn being avoided.
- head's `max spins` occasionally exceeds base's (7,616 / 14,848 in the loaded series). Counter
  artifact, not a regression: head accumulates across yields, so a long head wait is 100+
  yields with at most **64 contiguous** spins between them. Contiguous burn is bounded by
  construction on head; every base spin is contiguous.

`park_not_spin_lockstats_2026-08-09.tsv` (n=5 interleaved) is **TAGGED LOADED** — a foreign
100%-CPU process held one core throughout. Its ratios are usable (0.949 at 8bpc t=16, matching
the idle sweep's 0.955); its absolute ms are not.

### Correctness

- Corpus **766/766**, `md5_inventory` set-diff **BY NAME with the actual md5 as the value**
  against `aarch64_md5_fixes_2026-08-07_final.tsv.zst`: only-in-baseline 0, only-in-head 0,
  **differing 0**.
- Output **byte-identical base vs head, 10/10 cells** (both vectors × t=1/2/4/8/16).
- `cargo test --release -p rav1d-disjoint-mut` green; `wide_exclusion` green under
  `--features __probe_wide`. `cargo fmt --all --check` clean; `cargo check --release
  --all-targets` clean.
- **`mt_stress` 1/2/4/8/16 ×5 PASS**; `multi_decoder_pressure.sh` 12 procs × 3 iters ×
  threads 1/2/4/8/16 **PASS**, every md5 matching the serial reference.
- **Instrument verified against the binary, not the source.** `otool -tvV` on
  `TinyLock::lock_slow`: base is a frameless leaf looping on `isb`/`ldrb` with no exit; head
  has `cmp w8,#0x3e` + `bl _sched_yield`. `lock`/`try_lock`/`unlock` are unchanged single
  atomics in both, so the frameless leaf `add` is intact. head's `lock_slow` does gain an
  `stp x29,x30` frame — worth noting, but it is the cold path.
- The probe additions are **code-inert in the default build**: `__text` is byte-identical
  between the rebuilt branch head and the binary the sweep actually ran.

### Not measured — do not extrapolate

- **x86_64.** Nothing here transfers: `spin_loop()` is `pause` there at a different cost, and
  the scheduler is not heterogeneous.
- **The oversubscription ladder.** t=12 (== logical cores), t=24, t=32 were queued and never
  got an idle window — two other campaigns held the box for the rest of the session. Whether
  the t=16 win grows, saturates or reverses past 16 is **unknown**, and "more threads than
  cores" is inferred from **one** oversubscribed cell plus the lock counters, not a ladder.
- **A per-decode distribution.** `scripts/perf/tail_sweep.sh` (new, samples every decode
  individually — the right instrument for a rare event, since 16 two-point rounds is only 16
  tail samples) never ran, same reason. The tail conclusion rests on 16 rounds per cell: enough
  to say the base band is ~8% and not ~60%, **not** enough to characterise a 1-in-1000 event.
- Any vector but `v4k_8tile` / `v4k_8tile_10b`, in particular nothing with loop restoration
  live. `--features unchecked`, Miri. **Any spin bound other than 64** — it was not swept; it
  is just the value the deleted probe used.

### Recommendation

Keep it, but re-title the claim. It is not a tail fix and it is not a t=8 fix; it is a
**~4.5% win at oversubscription** with everything else null and no correctness cost, plus the
removal of a genuinely unbounded spin whose worst observed episode was 110 µs. The index entry
"`TinyLock` backoff/yield — null, measured twice" should become "**null at t≤8 on medians AND
on tails; 0.955/0.960 at t=16 (threads > cores), measured 2026-08-09**" so the next agent does
not re-run either half.

The source doc comment on `lock_slow` has been rewritten to say all of the above; it previously
carried the wrong contention counter and the unsourced bands.

lilith:
## Correction to this PR's own body: the band evidence I cited is not in any committed record

The PR body justifies this change with *"8bpc t=8 bands of [67.8..114.6] and [73.2..112.8] around medians near 70"*. An independent verifier grepped for those figures and could not find them; I re-checked and **they appear in no committed benchmark file**. The only near-hits (`67.82/67.84/67.86` in `tracker_borrowcost_2026-08-08.tsv`) are unrelated 10bpc t=2 rows.

They came from an agent's report text, not from a committed record. Quoting a number into a PR body without a file behind it is exactly the provenance failure this campaign has been trying to stamp out, and it was mine. The motivating evidence for this change should be treated as **unsourced** until someone re-measures base and commits the bands.

## What the measurement actually shows

Two independent A/Bs now exist and they agree on direction, disagree on significance:

| | t=16 8bpc park/base | t=16 10bpc | bands |
|---|---|---|---|
| authoring agent, n=16, strict idle gate | 0.955 | — | **disjoint** |
| independent verifier, n=7 | 0.9571 | 0.9480 | **every cell OVERLAPS** (incl. idle-only subset: 0.9687 / 0.9472) |

The **ratios agree to three digits**, so the direction is real and consistent. The verifier's n and box conditions were worse, so this is a failure to confirm rather than a refutation — but as it stands **band-disjointness is not established**, and the ~4-5% at t=16 is the only cell where anything shows.

Also corrected by the verifier, and it undercuts a supporting argument I made: **`wide_probe::N_SLOW` is not a contention counter.** Its own in-source doc says it counts "`add_slow` entries (poisoned / a live wide record / multi-block)". So the "~0.02% contention rate" I derived from it does not mean what I said it means. The reasoning about heterogeneous cores and preempted holders stands on its own; the rate does not.

## What is not in dispute

`lock_slow` really did spin unbounded in every shipping build — the `yield_now()` was behind `__probe_lock_backoff`, which is not in `default`. Whatever the size of the win, an unbounded spin with no yield is not a defensible thing to ship, and the fast path is untouched by construction (`lock`/`try_lock`/`unlock` remain one atomic each; the yield lives only in the already-`#[cold]` slow path).

## What is owed before merge

The oversubscription ladder (t=12 == logical cores, t=24, t=32) **never ran** — it was scripted and queued, but two other campaigns held the box continuously. That ladder is where the effect should be largest and it is the measurement that would settle this. Keeping this open until it runs on a quiet box.

lilith:
## LEFT OPEN — deliberately. Records landed; the lever is unrefuted at t > cores

Triage of the 15 open PRs, 2026-08-13. **This is the only one of the fifteen still open, and the
reason is that the brief I was given about it was wrong.**

I was told this was "one of the four waiting-policy arms #504 measured as NULL". Reading the
committed data says otherwise, and the difference is a thread count.

### #504 did not measure this cell, and says so itself

| | cell | threads | n | result |
|---|---|---|---|---|
| **#504** | `c256x2048` | **t=8** | 15 | 4 arms, all inside a ±1.8% floor vs an in-grid identity control |
| **this PR** | `v4k_8tile`, `v4k_8tile_10b` | **t=1..16** | 16 | null at t=1/2/4/8 — **agreeing** with #504 — and **0.955 (8bpc) / 0.960 (10bpc) at t=16** |

`benchmarks/c256_contention_2026-08-11.meta:128`, #504's own not-measured list, ends with `t=16`.
So the two results do not conflict; one covers a regime the other explicitly did not.

t=16 is the only cell on this host where threads exceed cores (M4 Pro, 8 P + 4 E = 12 logical). On an
idle box with `foreign_max = 0` on every committed row, **p90 and max move by the same factor as the
median** — the whole distribution shifted rather than a tail being clipped — and the mechanism
counters agree: the 64-spin bound is reached 2,657-3,638 times per 6-frame run (so the change is not
inert), the worst single base acquisition burns **~110 µs at t=16** against ~11.5 µs at t=8, and
aggregate spin work falls ~23% = 0.64 ms/frame of CPU against a **3.3 ms/frame wall** win — meaning
most of the gain is **the holder getting a core back**, not the waiter's own burn being avoided.

`AGENT_BRIEF.md` §6 now carries two rows instead of one: the existing waiting-policy row is re-titled
**"CLOSED AT t <= 8"** with its scope stated, and a new row records this as **OPEN under
oversubscription** with these numbers and the settling measurement.

### Your records landed (#518, merged `71a0fbd`)

All four `benchmarks/park_not_spin_*` files, with a `LANDED 2026-08-13` section appended that scopes
them against #504 and does not touch a line above it.
`git merge-base --is-ancestor fe33cff origin/main` passes.

**And your contention count corrected `main` in two places.** Both
`crates/rav1d-disjoint-mut/src/tracker_shard.rs:582` and `docs/C256_CONTENTION.md` §5 said the earlier
TinyLock nulls were taken "where contention is ~0.02% of registrations". As this record establishes,
that is `wide_probe::N_SLOW` — a count of `BorrowTracker::add_slow`, the poisoned / wide-record /
multi-block path — divided by the registration population and quoted as a contention rate; its
provenance traces to an `add_slow` count at `benchmarks/verify_compose2_2026-08-08.meta:278`. Directly
counted: 45,401,450 acquisitions/frame, ~27,800 contended = **0.061% of acquisitions / 0.122% of
registrations**. Six times, against a different denominator. Both sites now carry a dated correction
pointing here. #504's 0.264% for `c256x2048` t=8 is a different cell, counted directly, and is
explicitly not disturbed — nor is §5's "at most 10.7% is waiting" arithmetic.

Also preserved: your two independent prices for `spin_loop()` (**8.28 ns** here, 7.6 ns in #504 two
days later — both kept and dated, neither being the ~627 ns that applies under real contention), and
your **self-refutation**, which is the rarest thing in this whole set:

> The quoted 8bpc t=8 bands [67.8..114.6] and [73.2..112.8] appear in no committed benchmark TSV in
> this repo … and in no issue or PR body except #471 itself.

That is this PR's own body being checked and found unsourced, by the round that wrote the PR. Base
band widths are ~8% of the median, not the ~60% the rationale assumed, so **judged on the criterion
this PR asked for — p90, max, band width at t=4/8 — the change is null.** It survives on a different
result than the one it predicted. Third instance of the unsourced-number class the campaign has had
to correct.

### What would settle it — and what I did NOT do

**Not re-measured.** Another agent holds the box for timed work in a sibling repo; a contended-lock
A/B under foreign load would be exactly the wrong thing to trust, which is what this PR said about
itself on 2026-08-09 and is still true.

To close it either way:

1. **The t=12 / t=24 / t=32 ladder** — queued on this branch and never got an idle window. Whether
   the t=16 win grows, saturates or reverses past 16 is unknown, and the "more threads than cores"
   mechanism is inferred from a **single** oversubscribed cell plus the lock counters, not from a
   ladder. Judge it against a same-source control, not against `main`'s binary (#506).
2. **x86_64** — nothing here transfers: `spin_loop()` is `pause` there with a different cost, and the
   scheduler is not heterogeneous.
3. **A per-decode distribution** — `scripts/perf/tail_sweep.sh` exists on this branch and never ran.
   16 two-point rounds is 16 tail samples: enough to say the base band is ~8% and not ~60%, not
   enough to characterise a 1-in-1000 event.

Do **not** re-litigate this at t <= 8. That is now null on two cells and two vectors, and re-running
it there is the third repetition of a settled negative.

The branch is stale against `main` (`tracker_shard.rs` has moved substantially since `5807083b`), so
whoever picks this up should rebuild the arm on current `main` rather than rebasing this diff — the
four `__probe_lock_*` features #504 added already provide the backoff arm, so the code cost is close
to zero and the whole remaining task is the ladder.

Indexed in `benchmarks/pr_triage_2026-08-13.meta`.

PR #472, opened 2026-08-09
fix(soundness): the rectangle guard hands out ROWS, not the hull — same one registration, no overlapping `&mut`

lilith:
Base `perf/strided-rect` @ c1355c4 (PR #469, draft/blocked). Record:
`benchmarks/rect_rowview_2026-08-09.meta` + `rect_rowview_gap_2026-08-09.tsv` +
`rect_rowview_bands_2026-08-09.txt`.

## The defect, and why the fix is a reference change and not a tracker change

`index_rect_mut` registered the exact rectangle and returned a `&mut [T]` over the **hull**.
#469's CHANGELOG stated that split on purpose — *"records the hull as the reference and the
exact rectangle as the record"* — and that is precisely the bug: **the reference is what
Rust's aliasing rules bind to, not the record.** The two tile columns the rectangle record
exists to *accept* were then two simultaneously-live overlapping `&mut`.

`DisjointMutRect` / `DisjointImmutRect` keep the ONE registration and hand out one **row** at
a time. No reference is created at construction — the guard keeps the raw base pointer the
old code already derived from `get_mut` — so every reference is a subset of what the tracker
reserved, and two guards from different tile columns produce disjoint references. `row_mut`
takes `&mut self`, so two rows of one guard cannot be live at once either. **No extra atomic,
no extra lock**: the reference shape and the record shape are independent, and only the record
costs a registration.

## What it costs — two of the seven call sites could not be converted

`for_rows`, `for_rows_mut`, `compact_read_per_row`, `compact_write_back_per_row` and
`LfBlock::fill_hull` index by row and convert mechanically.

`narrow_guard` / `narrow_guard_mut` hand the caller a FLAT slice over the hull plus a stride.
That shape *is* a hull reference; narrowing the record does not narrow it. Both revert to a
plain hull borrow and both callers revert to gating on `tile_threading_active()`:

- **`with_pixel_guard_immut`** — the condition #469 changed sits two lines under a doc comment,
  untouched by that diff, which already said wide guards *"are unsafe under tile threading even
  in the immutable case"*.
- **`block_mut`** — loses #469's copy elision under tile threading and takes the compact path
  again. The registration win still reaches it one level down: the compact copies reserve the
  block ONCE, so a block costs **2** registrations instead of `2h` (`main` pays `2h` and the
  same two memcpys). Making that site zero-copy AND sound needs `BlockMut::as_mut_bytes` to
  become a row API, which is ~27 kernel call sites in `itx_arm` / `mc_arm` / the wasm files plus
  the ~40 `with_pixel_guard_mut` closures — named, not attempted.

## Perf — the win survives at every t > 1, at about two thirds of #469's

n=7, 224 rows, four arms interleaved with rotating order inside every cell,
two-point wall fit at 2 and 20 frames, dav1d 1.5.4 `--framedelay 1` in the same
sweep. **The box was NOT idle** — two other agents were measuring and building on
it for the whole window, and the strict idle gate could not commit a single cell in
20 minutes, so `ALLOW_LOAD=1` was used and all 224 rows carry their foreign-process
count. Absolute ms are inflated; the arms all see the same load, so the paired
ratios are the result.

| vector | t | base | #469 | head | dav1d | head/base | #469/base | retained | head/ref | #469/ref |
|---|---|---|---|---|---|---|---|---|---|---|
| v4k_8tile | 1 | 321.6 | 326.8 | 327.1 | 249.1 | 1.0171 | 1.0162 | 105% | 1.313 | 1.312 |
| v4k_8tile | 2 | 201.5 | 169.4 | 180.6 | 127.2 | **0.8963** | 0.8406 | 65% | 1.420 | 1.332 |
| v4k_8tile | 4 | 117.3 | 92.7 | 99.8 | 67.6 | **0.8508** | 0.7901 | 71% | 1.477 | 1.372 |
| v4k_8tile | 8 | 74.8 | 60.0 | 66.4 | 39.8 | **0.8878** | 0.8024 | 57% | 1.667 | 1.506 |
| v4k_8tile_10b | 1 | 363.9 | 367.4 | 365.7 | 253.9 | 1.0049 | 1.0096 | 51% | 1.440 | 1.447 |
| v4k_8tile_10b | 2 | 211.3 | 192.4 | 195.0 | 129.5 | 0.9227 | 0.9104 | 86% | 1.506 | 1.486 |
| v4k_8tile_10b | 4 | 123.4 | 105.8 | 110.8 | 68.9 | 0.8982 | 0.8573 | 71% | 1.609 | 1.535 |
| v4k_8tile_10b | 8 | 79.8 | 68.8 | 70.3 | 40.7 | 0.8809 | 0.8621 | 86% | 1.728 | 1.691 |

`retained` = `(base - head) / (base - #469)`, i.e. how much of #469's distance from
`main` this keeps. Scaling t1→t8: 8bpc **4.30x → 4.93x** (#469 5.45x, dav1d 6.25x);
10bpc 4.56x → 5.20x (#469 5.34x, 6.24x). head-vs-base bands are disjoint at 6 of 8
cells; the two OVERLAP cells are 10bpc low-thread, where the load made them wide.

**The missing third is `block_mut`'s copy elision — exactly the part that needed the
hull reference.** Nothing else differs between the two arms. **t=1 is unchanged by the
row view**: 1.017 against #469's 1.016 is the same number, so that cost belongs to the
tracker change, not the guard. And the 1.30x bar is still met at no cell.

### The t=1 fix #469 names is a MEASURED NULL

Built exactly as described — a `COLS` const parameter on `ShardRecs::find`/`::alloc`
selected at `add`'s fast path by `mask != 0`, skipping both the `cols_meet` test and
the two column stores, plus a reset of every `c0`/`c1` pair in `reprovision` (a later
`mask` 0→non-zero flip reading a stale narrow range would be a MISSED overlap, not a
false positive). The identity it rests on is exact: on a one-shard instance every
registration carries `0 ..= COL_ANY`, and `cols_meet(a0, a1, 0, COL_ANY)` is
unconditionally true. All `rav1d-disjoint-mut` tests pass with it. A/B at n=9 on three
cells, near-idle (2 of 27 rows tagged):

| cell | cols | head | ratio | bands |
|---|---|---|---|---|
| 8bpc t=1 | 324.6 [323.6..337.3] | 325.7 [323.3..334.5] | 0.9966 | OVERLAP |
| 8bpc t=8 | 56.6 [54.9..83.9] | 55.8 [54.6..83.1] | 1.0139 | OVERLAP |
| 10bpc t=1 | 363.7 [361.8..367.2] | 362.3 [360.5..378.4] | 1.0037 | OVERLAP |

Every band overlaps and the sign is inconsistent. **Reverted** (`rect_rowview_cols_ab_2026-08-09.tsv`
keeps the rows). #469's own "0.991 with the two column stores deleted" does not
reproduce as a real effect either — whatever the t=1 cost is, it is not the column
fields, and the next attempt should profile before it patches.

## Gates

- **Miri**, one test at a time, **9/9 pass** including the two that were UB.
- **Teeth measured, not assumed.** Planting the *transient* hull reference back inside
  `row_mut` (rebuild `from_raw_parts_mut(base, (h-1)*stride + w)`, index the row out of it)
  leaves `writing_through_both_guards_while_both_are_live` and
  `four_tile_columns_the_way_the_decoder_holds_them` **GREEN** — the two cases that originally
  caught the defect are not sufficient once the API is row-shaped, because each of their row
  references dies at the end of its own statement. Two new cases hold a row reference ACROSS
  the next guard's retag and do catch it, plus a non-Miri liveness case asserting the rows
  handed out really are the strided rectangle. Mutation restored byte-exact (sha256 + `git diff`).
- **Corpus 766/766**, set-diff BY NAME with the md5 in the value against
  `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`: 0 only-in-baseline / 0 only-in-head /
  **0 differing**.
- Output md5 identical across `main` / #469 / this branch on 5 vectors × t=1/2/4/8/16 (25 cells,
  all three arms agree, thread-invariant).
- `cargo test --release --workspace`: **254 passed, 0 failed**. `cargo fmt --check` clean;
  `cargo clippy -p rav1d-disjoint-mut --all-targets -D warnings` clean; `cargo doc -p
  rav1d-disjoint-mut` with `-D warnings` clean.
- Oracle soak `RECT_ORACLE_ITERS=200000` reproduces #470's committed numbers exactly.
- Both standing hazards re-planted under `--features __probe_wide` (4af62ae's in-lock `state`
  re-read deleted; `active()` cut to one shard): `wide_exclusion` **FAILS 3/3** each, restored
  byte-exact.
- `#![forbid(unsafe_code)]` proved ACTIVELY on `rav1d-safe` (planted `unsafe` → the lint errors
  at the attribute; restored → green).
- `mt_stress_4k` (thread counts 1/2/4/8/16 × 5 trials) ok; `multi_decoder_pressure.sh` 12
  concurrent decoder processes × 3 iters × {1,2,4,8,16} over 5 vectors — **PASS**, every md5
  matching the serial reference, no wedge, no overlap panic.

## CI breakage on #469 that this also clears — and why CI cannot show it here

`ci.yml` triggers on `pull_request: branches: [main]`, so a PR based on `perf/strided-rect`
runs ONLY the `disjoint-mut CI` workflow. Clippy, the two i686 legs, `wasm32-wasip1` and
`macos-15-intel` never fire on #472. The 32-bit fix is therefore A/B'd locally instead —
and `cargo build -p <lib>` of an rlib needs no C linker, which is the cheap way to reach the
MIR pass `cargo check` stops short of:

| target | c1355c4 (#469) | this branch |
|---|---|---|
| `i686-unknown-linux-gnu` | `error: this arithmetic operation will overflow` (tracker_shard.rs:1294) | builds |
| `wasm32-wasip1` | same error | builds |

- The two clippy findings and the three unresolved intra-doc links: `Lint` and `Documentation`
  are **green** on #472 and were red on #469, so those two the workflow does show.

## What this does NOT fix — read before merging

The macOS-Intel job can still fail on this branch, and it is **not** the aliasing bug.
A fresh failure on the merge commit c1355c4 (run 31294761102, job 93197936544) carries a
full stack that #469's verification round did not have:

```
thread 'rav1d-worker-3' panicked at include/dav1d/picture.rs:568:19:
  overlapping DisjointMut:
   current:    & _[241..16385] on ThreadId(22)
  existing: &mut _[16384..20240]
   3: add_wide<false>          tracker_shard.rs:2286
   4: add_rect_immut           tracker_shard.rs:1709
   8: compact_read_per_row     picture.rs:1254
   9: loopfilter_sb_dispatch   src/safe_simd/loopfilter.rs:5134
  14: rav1d_filter_sbrow_deblock_cols
```

Two things follow **without knowing the plane stride**:

1. The last byte of a strided rectangle's hull is by construction the last element of its last
   ROW, never a gap. A's hull ends at 16385, so A covers byte 16384; B covers `[16384, 20240)`.
   **The two records' covered byte sets intersect at 16384 — a genuine overlap, not a `COL_ANY`
   false positive and not an artefact of the hull reference.**
2. `loopfilter_sb_dispatch` in `src/safe_simd/loopfilter.rs` is `#[cfg(target_arch = "x86_64")]`
   (wasm32 shares it). aarch64 dispatches to `loopfilter_arm`, which never calls
   `compact_read_per_row`. **That is why 1,600 aarch64 runs could not reproduce it: the failing
   call site is not compiled on that box.**

What #469 changed is the WINDOW, not the extent. `main` reads the same tap window through `h`
per-row guards, each live only for its own memcpy; the rectangle reserves all `h` rows for the
whole copy loop. The bytes are identical — so `main` has the same unsynchronized overlap and
only detects it when the conflict happens to land inside the one row whose guard is live at that
instant. **This is a property of "one registration for a multi-row access", not of the hull
reference, so the row view does not remove it.** Repro rate as measured, n=3: 1 of 2
macos-15-intel runs on 4db4144, 1 of 1 on c1355c4, 0 of 1,600 aarch64 runs at a site not
compiled there.

The overlapping column is the loop filter's deliberate over-read: the 8bpc H-luma window is
`tap_before + tap_after = 7 + 9 = 16` columns because the kernel loads 16 B/row, while the tap
ladder only reaches p6..q6 — columns +7 and +8 are read and never used. That same file already
documents narrowing this window once for exactly this reason. Fixing it is an x86-only change I
could not run here (no Rosetta on this box), so it is written up rather than guessed at:
`benchmarks/rect_rowview_2026-08-09.meta` §3.

## Exactness

#469's body says the test is "EXACT, not conservative". It is exact only between two records
that BOTH carry a real column range; `ShardRecs::c0`'s own doc was right and the body was not.
A row-crossing interval registers `0..=COL_ANY` and degrades to the plain hull test. Re-measured
here at `RECT_ORACLE_ITERS=200000` — 1,600,000 pairs, **0 missed overlaps**, 2,163 conservative
false positives (0.14% of disjoint pairs), every one involving a row-crossing interval. The doc
now says that, says why the number is a budget (a false positive is a decode failure), and
records the exact fix that was deliberately NOT taken: a `contig` bit plus a three-probe test has
no free home in the 128-byte record (`SLOTS` already went 7 → 5 to fit `c0`/`c1`), and a wrong
three-probe test yields a MISSED overlap — the direction that must never regress for a
false-positive win.

## Not done

- **The t=1 regression is not fixed** — the fix #469 names was built and measured NULL
  (above), so it is reverted rather than shipped. The cost is still there and still
  unattributed.
- **Per-call-site registration counts** (`--features probe-sites`) were not re-measured for this
  arm; the count claim here is structural (`2h` → 2 per block) rather than measured.
- x86_64: nothing built or run (no Rosetta on this box) beyond the CI evidence above.
- Loop restoration is off in both 4K gap vectors, so `looprestoration*.rs` registers zero borrows
  in this grid.
- Negative strides: every rectangle caller falls back, so that path is unexercised here.


lilith:
### CI on this PR, and what it can and cannot say

`ci.yml` triggers on `pull_request: branches: [main]`, so a PR based on `perf/strided-rect` runs **only** the `disjoint-mut CI` workflow — no Clippy, no i686, no `wasm32-wasip1`, no `macos-15-intel`. What did run:

| | #469 (c1355c4) | here (b2b39e3) | main (ee07b00) |
|---|---|---|---|
| Documentation | **fail** | **pass** | pass |
| Lint | **fail** | **pass** | pass |
| MSRV (1.85) | pass | pass | pass |
| Wide-path exclusion gate | pass | pass | pass |
| Test (macos/ubuntu/windows, default) | pass | pass | pass |
| Miri (Object) | fail | fail | **fail** |
| Test ×4 (`--no-default-features` / `--all-features`) | fail | fail | **fail** |

The five red legs are **byte-for-byte the same set that is red on `main`** — `cannot find module or crate `std``, i.e. the test harness under `--no-default-features`. Nothing new is red here, and Documentation and Lint are repaired.

The two legs that CI cannot run were A/B'd by hand instead. `cargo build -p rav1d-disjoint-mut --lib` of an rlib needs no C linker and still reaches the MIR pass `cargo check` stops short of:

| target | c1355c4 | b2b39e3 |
|---|---|---|
| `i686-unknown-linux-gnu` | `error: this arithmetic operation will overflow` (tracker_shard.rs:1294) | builds |
| `wasm32-wasip1` | same error | builds |

`macos-15-intel` is the one that matters and the one I could not reach — no Rosetta on this box. See the PR body: that failure is a **genuine** byte overlap on an x86-only call site, and the row view does not remove it.

lilith:
## Closing: BOTH halves landed — the row view via #505/#513, the x86 overlap via #497

Triage of the 15 open PRs, 2026-08-13. This is the one supersession claim in my brief that survived
checking intact, and it survived on both of its halves.

### Half 1 — "the guard hands out ROWS, not the hull" is `main`'s default path

`crates/rav1d-disjoint-mut/src/lib.rs` now has `DisjointImmutRectGuard` / `DisjointMutRectGuard`, and
the doc on the latter states this PR's argument as the invariant:

> [`Self::row_mut`] takes `&mut self`, so **at most one row reference exists** …

`row_mut(&mut self, r: usize) -> &mut [V]` is at `:1615`; the ONE registration is
`add_rect_mut` / `add_rect_immut`. #505 built it, and **#513 flipped it to the DEFAULT path on
2026-08-11** — `Cargo.toml` carries `__lf_rect` and `__rows_rect` as declared no-ops saying exactly
that. Measured on the way in: −2.0% to −2.5% wall at t=8 on the 1024-wide family from the CDEF seam,
−3.9% to −4.5% composed with `LfBlock::fill`'s (`docs/LAYOUT_CDEF.md` §4).

### Half 2 — the `macos-15-intel` panic was real, was on `main`, and is fixed

This PR was careful to say the overlap it hit was **not** an artefact of its own change:

> `main` reads the same tap window through `h` per-row guards … so `main` has the same unsynchronized
> overlap and only detects it when the conflict happens to land inside the one row whose guard is
> live at that instant.

That reading was correct. It was independently found, root-caused and fixed as the **#494 x86_64 t=8
race**, recorded at `docs/X64_APPLICABILITY.md` §G. The named counterparty matches this PR's stack —
`loopfilter.rs:5134:44`, the V-run compact read, against `owned_recon.rs:937:42` and against the
row's own `DeblockCols`. Root cause: the x86 dispatcher sized the V window from a **constant**
`tap_before = tap_after = 7` instead of the mask-selected reach, reading 3 rows past its own
superblock row. Fixed by `lf_run_reach()` (#497), and it is now a **deterministic single-threaded**
`debug_assert!` in `loopfilter_sb_direct` rather than a rare t=8 race.

This PR also flagged the right general question — *would ONE exact record ever reject what `h`
per-row records permit?* — and `main` answered it: `bounds_probe::eval_rect` measures **`rect_ovl = 0`
at every site** over the corpus, i.e. every hull collision is a gap collision and the exact rectangle
never collided once (`BOUNDS_MAP.md`, "The strided-rectangle counterfactual"). Your `ShardRecs::c0`
correction — that the test is exact only between two records that both carry a real column range —
is preserved there too.

### What is not on `main`

`rect_rowview_*_2026-08-09.*`, including the `cols_ab` A/B (0.9966 / 1.0139 / 1.0037, every band
overlapping, reverted). That question has since been answered from a different direction and the
answer is not "the column stores": **#506 measured the t=1 cost as CODE PLACEMENT** — 4,828 B of
provably-dead `#[used]` text with 0 symbols resized costs +1.10% at t=1, 0 of 11, and against a
same-source control the rectangle costs **0.9967 (7/9)**. Reachable at `pull/472/head` =
**`b2b39e3f86ea412ed123a4511f24669d9ccc3bcb`**; indexed in `benchmarks/pr_triage_2026-08-13.meta`.

PR #473, opened 2026-08-09
perf(tracker): tile-keyed borrows — Variant 2 measures a negative, Variant 3 (the hull the key makes legal) is 0.83x at 8bpc t=8, neither beats #469

lilith:
Prototypes and prices a **tile-keyed** borrow system for issue #455, in the three
variants the task named. Base `main` @ `ee07b00`. **Not for merge into
`default`** — every behavioural change is behind a `probe-*` / `__probe_*`
feature, and §"Why this is not mergeable" says exactly why.

Full record: `docs/TILE_KEYED_BORROWS.md` + `benchmarks/tile_keyed_2026-08-09.meta`.

## The idea, in one paragraph

The tracker shards by ADDRESS, and that one choice has to serve two requirements
that pull in opposite directions — a fine block shift separates two tile columns
on the same picture rows but makes a strided access pay a shard line per row; a
coarse one keeps the access on one line but puts every tile column on it. The
`BLOCK_SHIFT` ladder is the record of that being decided in both directions.
Both requirements are about TILES, and AV1 already guarantees tile regions do not
overlap during reconstruction. So put the tile on the borrow.

## Result first, because it is mixed

Idle box, strict gate, **`foreign = 0` on every committed row**, dav1d 1.5.4
`--framedelay 1` in the same interleaved sweep with rotating arm order,
two-point wall fit at 2 and 20 frames. **n = 4 rounds at 8bpc, n = 3 at 10bpc —
below this campaign's n >= 7 bar**, because the box was shared with another
agent's test suite for most of the window and the strict gate discarded rather
than commit loaded cells. Nothing below should be read as a settled three-digit
figure.

| 8bpc | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| base/dav1d | 1.303 | 1.580 | 1.631 | 1.860 |
| **tkboth** (shard + hull) | 1.329 | **1.401** | **1.385** | **1.541** |
| tkshard (shard only) | 1.340 | 1.634 | 1.625 | 1.806 |
| untracked ceiling | 1.160 | 1.264 | 1.262 | 1.345 |

`tkboth/base` is 1.0203 (OVERLAP) / **0.8868** / **0.8490** / **0.8287**, the
three t>1 cells disjoint. 10bpc: 1.0785 / 0.9939 / 0.9693 / 0.9285, taking
1.453 / 1.649 / 1.681 / 1.900 to 1.567 / 1.639 / 1.629 / 1.764.

**The ~1.30x bar is met at no cell**, and `tkboth` is BEHIND #469 everywhere:
that branch's own verification round reports 1.309 / 1.332 / 1.323 / 1.477 at
these 8bpc cells (quoted from its record, not re-measured here). Tile identity
is a *cleaner* key than an address-derived column band — exact, no Lemire magic,
no per-`alloc` column stores — and on this vector a *slower* one.

**The two halves behave completely differently, and only one of them is the
money.** `tkshard` — Variant 2 exactly as the task framed it, "one lock per tile index",
with references untouched — is a **measured negative**: 1.0285 / 1.0338 /
0.9964 / 0.9712 at 8bpc and 1.07 / 1.07 / 1.05 / 1.02 at 10bpc, the t=1/t=2
cells disjoint-WORSE. With 128 shards the existing multiplicative address hash
already separates concurrent tile columns well enough that naming the tile buys
nothing measurable, while the key's own store and compare cost 2-7% where there
is nothing to win. That is the `TinyLock`-backoff shape a third time: a
plausible contention fix the contention does not justify.

So essentially the whole 8bpc t>1 win is the HULL, and **the key's value is as
an ENABLER, not as a shard selector.** The hull without the key does not decode
at all — `overlapping DisjointMut: current &mut _[960..58576] existing &mut
_[0..57616]`, two 16x16 hulls at pixel columns 0 and 960, the tile-column
boundary at 3840/4, which is #467's documented false positive reproduced.

Registrations/frame at t=8 (`probe-tilekey-count`): **26,484,179 -> 15,540,586
(-41.3%)**, the keyed population itself 13,212,787 -> 2,269,194 (-82.8%). The
count reduction and the deleted per-row compact copies are **not separated** —
the hull removes both at once and no arm here isolates them.

## What each variant cost

**Variant 1 — separate owned buffer per tile.** Reachable in principle and
*not* blocked by #468's objection, which is about the SHARED picture: in the
default build the inner type is `PicBuf`, not a raw pointer
(`include/dav1d/picture.rs:406`), and `wrap_buf` (`:560`) already builds a
component from an owned buffer. **Priced, not built** — `examples/tile_stitch_cost.rs`
on the real `v4k_8tile` geometry, median of 9:

| arm | ms/frame | band |
|---|---|---|
| `memcpy_flat` (contiguous 12.44 MB, lower bound) | 0.170 | [0.168..0.629] |
| `stitch_sbrow` (34 bands x tiles — the shape a decoder needs) | **0.315** | [0.304..0.351] |
| `write_into_pic` | 2.553 | [2.434..2.612] |
| `write_into_tile` | 2.567 | [2.400..2.576] |

The stitch is affordable — 0.315 ms/frame, 1.6% of the tracker's 19.7 ms at t=8
— and it does not get worse per-sbrow. **But the cache-locality bonus does not
exist**: writing recon into a private compact tile buffer measures 2.567 against
2.553 into the plane, bands overlapping. Variant 1 is `write_into_tile + stitch`
against `write_into_pic` alone, i.e. strictly +0.315 ms of extra work.

What blocks it is **coordinates, not ownership**: 22 sites do
`4 * (t.b.y * pixel_stride + t.b.x)` in FRAME coordinates plus ~10 reads of
`f.cur.stride[..]`, and a tight-stride tile buffer needs `x - tile_x0`,
`y - tile_y0` and a different stride at every one. The translation-free variant
(full-width buffer per tile) costs 4x memory for a tile row's columns, ~50 MB at
4K. Not a one-commit change.

**Variant 2 — one lock per tile index.** Built, measured, negative (above).

**Variant 3 — precomputed per-tile row refs.** Subsumed by the stronger move the
key makes available: one strided-hull registration instead of `h` per-row ones.
This is the win, and it composes with Variant 2 exactly as the task predicted —
just with Variant 2 contributing ~nothing.

## Why this is not mergeable

Two independent blockers, both named with the specific mechanism:

1. **Shard-by-key is unsound while the filter chain is unkeyed.** A keyed borrow
   lands in its tile's shard, an unkeyed one in an address-hashed shard; they are
   different partitions of the same array and never meet, so a genuine overlap
   between reconstruction and the filter chain would be missed. The filter chain
   is 6,389,542 registrations/frame at t=8, and routing that volume through the
   wide path is what made #469's avalanche 2.28x slower than base.
   `tests/tile_key.rs::keyed_versus_unkeyed_overlap_is_missed_and_that_is_the_open_edge`
   asserts this EXECUTABLY, so the day it starts detecting, the gate fails and
   says why. Closing it is scoped with file:line in the doc.

2. **The hull hands out a `&mut` over the hull.** Two tile columns' hulls on the
   same rows are two live overlapping `&mut`, which is UB by Rust's aliasing
   model whatever the tracker records — the same defect #470 found in #469, from
   the other direction (there the record was narrower than the reference; here
   the reference is wider than the bytes touched). The sound form needs the
   per-row-view guard #470 named. **So the hull arm's wall number is a ceiling
   from a build with UB in it, not a candidate.** The shard arm has no such
   problem: it keeps today's per-row references.

## Correctness and gates

* `md5_inventory` **766 PASS / 0 mismatch / 0 error / 2 skip**; set-diff BY NAME
  with the md5 as the value against
  `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`:
  **only in baseline 0 / only in head 0 / differing 0** (768 rows both sides).
* Cross-arm frame md5 identical over {main, base, tkshard, tkboth, untracked} x
  t = 1/2/4/8 x both depths — 40 cells, `a00c11f454328023c58af14d55544cff` and
  `4f218411bc6ee4cc9c630fe827337fa2`.
* `cargo test --release` **157 passed, 0 failed** (28 ignored), incl.
  `decode_md5_verify` 14/14, `decode_permutations` 19/19, `mt_stress`,
  `tile_threading_overlap`. `cargo fmt --all --check` clean.
* New mechanism gate `crates/rav1d-disjoint-mut/tests/tile_key.rs`, 6 tests,
  green WITH and WITHOUT the feature (it asserts the inert behaviour when off, so
  it cannot go vacuous by being built wrong). **Teeth by mutation x3** — key as a
  global off-switch, keyed twins dropping the key, shard index ignoring the key —
  3 of 6 tests fail on each, each restored by sha256 **and** `git diff --exit-code`.
* Both mandatory pre-existing hazards re-planted under `--features __probe_wide`:
  `add`'s in-lock `state` re-read deleted (4af62ae) -> `wide_exclusion` FAILS;
  `active()` cut to one shard -> FAILS. Restored byte-exact, green again.
* `#![forbid(unsafe_code)]` proved ACTIVELY (planted `unsafe` -> E0133).

## Three things worth carrying even if none of this ships

1. **A thread-local is not a free channel on macOS.** Measured +0.502 ns/call
   out-of-line (an argument and a global atomic are both +0.000), because
   `TLS.with(..)` compiles to an indirect `blr` into `_tlv_get_addr` plus the
   stack frame that forces. At 22.7 M registrations/frame that is 11.4 ms against
   a 19.7 ms tracker. **And the first micro-benchmark said +0.00 ns and was
   wrong** — LLVM hoists `_tlv_get_addr` out of a caller's loop, so the read has
   to be made in an out-of-line callee to be priced at all.
2. **The per-tile buffer's cache-locality bonus does not exist.** Writing recon
   into a private compact tile buffer measured 2.567 ms/frame against 2.553 into
   the shared plane — overlapping bands, null. The stitch is therefore pure
   added work (0.315 ms/frame), not partly paid for by locality.
3. **A cfg naming the sub-crate's feature inside the parent crate is a silent
   null.** The first hull build spelled `__probe_tilekey_hull` where the parent
   crate's feature is `probe-tilekey-hull`. It compiled, decoded correctly, and
   measured identically to base. The registration COUNT caught it — 26,484,179
   on both arms, to the byte. Assert on the mechanism, not the outcome.

## Not measured — before the wins

* **x86_64**, and this is the change most likely to differ there: the +0.502 ns
  thread-local penalty is a macOS `_tlv_get_addr` property and does NOT transfer
  to Linux/ELF initial-exec TLS, so on x86_64 the key channel may be free and the
  design's shape changes.
* **Variant 1 beyond its stitch** — no tile-buffer decoder was built.
* **Any vector with loop restoration live.** Unchanged structural blindness
  (#455 item 4): both 4K gap vectors are intra-only with LR off, so the part that
  is NOT keyed is understated here relative to a corpus vector.
* **Miri** on the hull arm; the doc says on paper why it would fail.
* t=16, `--features unchecked`, and every vector but `v4k_8tile{,_10b}` in the
  timed grid.
* **The remaining 3-4 sweep rounds — given up deliberately.** 168 rows, 21
  discards, foreign-load column summing to 0. When the other agent's suite
  finished it started its own TIMED sweep, so both were polling for an idle box
  and each was dirtying the other's cells; I stopped mine, because every
  discarded attempt is ~90 s of decoding loaded onto their measurement. Same
  conclusion the site-class round reached: two timed campaigns cannot share this
  box. Regenerate the tables with `scripts/perf/gap_bands.py` if anyone extends
  the TSV.

## Also worth carrying, from the carrier itself

`base` (this branch, default features — the wider `WithOffset` plus the recon
key stores, no tracker change) against pristine `main`: nothing measurable at
8bpc (all four bands overlap), but **+1.6% at 10bpc t=2 and +2.1% at t=4, both
disjoint**. `WithOffset` goes 16 -> 24 bytes in a `Copy` struct passed by value
through every kernel, and at two bytes per pixel that shows. Any future design
that puts a key on the reference pays this before it wins anything — which is
part of why 10bpc barely moves here.


lilith:
## Closing: its three verdicts are all on `main`, and the design it rejected is the one that shipped

Triage of the 15 open PRs, 2026-08-13. This round said up front it was **not for merge** ("every
behavioural change is behind a `probe-*` feature" + a §"Why this is not mergeable"), so closing it is
what it asked for. Its findings landed:

| finding | where it is on `main` |
|---|---|
| **Variant 2 (one lock per tile index) is a measured negative** — the 128-shard address hash already separates concurrent tile columns, and the key's own store/compare costs 2-7% where there is nothing to win | `OWNERSHIP_MODELS.md` §3, "Keying locks by tile — **unsound unless the keying is total**" — and #504 re-states the soundness half: "keying a record on its worker, refuted on soundness grounds (partial keying misses real overlaps)" |
| **the stitch is affordable — `stitch_sbrow` 0.315 ms/frame** | `OWNERSHIP_MODELS.md:120`, verbatim |
| **the per-tile buffer's cache-locality bonus does not exist** — `write_into_tile` 2.567 vs `write_into_pic` 2.553, bands overlapping | `OWNERSHIP_MODELS.md:127` |
| **the blocker is coordinates, not ownership** — 22 sites computing `4 * (t.b.y * pixel_stride + t.b.x)` in FRAME coordinates | `MUT_RECON_KERNELS.md:70` and `src/owned_recon.rs:698` both name it as "#473's 22-site coordinate cost" |

**And the coordinate blocker was paid.** `src/owned_recon.rs` (#482) is exactly the change this round
scoped and declined: a column-compact band with its **own** stride, `(&mut [u8], base, stride)`
kernel signatures carrying the origin. #481 stated the design rule this round's Variant 1 pricing
implied — "the shippable geometry is column-compact, and it is not a follow-up, it is the same
change" — and #482 wrote it. So Variant 1 was right about the shape and right that it was not a
one-commit change.

Both of the blockers this PR named for itself remain true of *this* branch and were dissolved rather
than fixed: shard-by-key is still unsound while the filter chain is unkeyed, and the hull arm still
hands out a `&mut` over the hull — the defect #470/#472 found from the other side, and which
`DisjointMutRectGuard::row_mut` (`lib.rs:1615`, default since #513) is the answer to.

### What is not on `main`

`docs/TILE_KEYED_BORROWS.md` (437 lines) and `benchmarks/tile_keyed_*_2026-08-09.*`, including two
lessons I could not find recorded anywhere on `main` and am therefore quoting here so they are at
least searchable from this thread:

* **a thread-local is not a free channel on macOS** — +0.502 ns/call measured out-of-line, because
  `TLS.with(..)` compiles to an indirect `blr` into `_tlv_get_addr`; and the FIRST microbenchmark
  said +0.00 ns and was wrong, because LLVM hoists `_tlv_get_addr` out of a caller's loop. This does
  **not** transfer to Linux/ELF initial-exec TLS, which is why the design's shape may differ on x86_64.
* **a cfg naming a sub-crate feature inside the parent crate is a silent null** — `__probe_tilekey_hull`
  vs the parent's `probe-tilekey-hull` compiled, decoded correctly, and measured identically to base.
  The registration COUNT caught it (26,484,179 on both arms, to the byte). Assert on the mechanism,
  not the outcome.

Reachable at `pull/473/head` = **`a2d4c822cbe6ff52ab67277d5c6e51f7bd8483bf`**; indexed in

Also worth recording that this round was honest about its own statistics: n=4 at 8bpc and n=3 at
10bpc, **below the campaign's n>=7 bar**, stated in its own result section rather than buried.

PR #474, opened 2026-08-09
perf(recon): owned per-tile reconstruction buffers (#455 Variant 1) — 8bpc t=8 1.876x -> 1.599x, free at t=1, but behind #469

lilith:
Builds **Variant 1** of issue #455's tile-keyed design — owned per-tile
reconstruction buffers with a per-superblock-row stitch — which PR #473 priced
but did not build. Base `main` @ `ee07b00`. **Not for merge**: behind
`--features tile-owned-recon` (default off) plus a runtime `RAV1D_TILE_OWNED`
switch, so both A/B arms are the same binary.

Record: `docs/TILE_OWNED_RECON.md` + `benchmarks/tile_owned_recon_2026-08-09.meta`
+ 3 data files.

## Result first, including where it loses

n = **9** rounds, 288 rows, **`foreign_max = 0` on every committed row**, dav1d
1.5.4 `--framedelay 1` in the same interleaved sweep with rotating arm order,
two-point wall fit at 2 and 20 frames, idle M4 Pro.

| 8bpc | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| base / dav1d | 1.293 | 1.585 | 1.625 | 1.876 |
| **tile-owned / dav1d** | 1.295 | **1.478** | **1.474** | **1.599** |
| tkon/base | 1.0010 | **0.9322** | **0.9068** | **0.8525** |
| bands | OVERLAP | disjoint | disjoint | disjoint |

10bpc: 1.429 / 1.586 / 1.580 / 1.722, head/base 0.9998 / 0.9764 / 0.9563 /
0.9065, the three t>1 cells disjoint. Scaling t1->t8 **4.68x -> 5.49x** at 8bpc
(dav1d 6.79x).

**The ~1.30x bar is met at no cell the change touches.** The only cell under it
is 8bpc t=1 at 1.295, which is base's number — the feature declines below two
workers.

**It does not beat #469.** That branch's own verification round reports
1.306 / 1.328 / 1.333 / 1.470 at these 8bpc cells (quoted, not re-measured
here) against this branch's 1.295 / 1.478 / 1.474 / 1.599. It IS ahead at t=1,
where #469 costs 1.4% and this costs nothing.

The arithmetic closes and says why: #469 removes the per-row split from BOTH
reconstruction and the filter chain (22.70 M -> 7.92 M registrations/frame,
−65%); this removes it from reconstruction only (22.70 M -> **13.37 M**, −41%).
At the campaign's measured 1.00 ns per removed registration, 9.33 M removed
predicts ~9.3 ms; measured is 68.2 -> 58.1 = **10.1 ms**. The count-is-cost rule
holds a third time. **The two are complementary, not alternatives** — private
recon buffers leave the filter chain as the only split population, which is
exactly what #469's rectangle record shrinks. Nobody has measured the
composition.

## The design, and why it is a three-line seam

Each tile gets its own plane set with the picture's **byte length and stride**,
so every frame-coordinate offset reconstruction computes indexes the same pixel
in the tile buffer as in the picture. The 22 `4 * (t.b.y * pixel_stride + t.b.x)`
sites and the ~10 `f.cur.stride[..]` reads that #473 named as the blocker are
untouched. The seam is the three `&f.cur.data...` acquisitions in `recon.rs`,
replaced by `Rav1dFrameData::recon_planes(tile_idx)`. `mc`'s acquisition stays
on `f.cur` deliberately — it is only the `ref_eq` intrabc identity test.

A per-instance `Rav1dPictureDataComponent::needs_row_split()` replaces the
process-global `tile_threading_active()` at the six branch sites that pick
between one hull borrow and `h` per-row ones. A privately-owned component
answers false, so reconstruction keeps the single-guard no-copy path at every
thread count; the shared picture is unchanged.

`stitch_sbrow` runs before the tile's progress counter is published, so the
filter task — gated on every tile having published that row — never sees a
partial row.

**No `unsafe` was written.** Ownership is static; borrowck and the existing
tracker both hold. Neither of #473's two blockers applies: there is no
keyed-vs-unkeyed shard partition (a tile buffer is a different `DisjointMut`
instance the filter chain never touches), and the hull a tile takes covers only
its own buffer, so the reference is as exclusive as the record.

## Costs, measured

* **Peak RSS 106.3 -> 202.4 MB** at 8bpc t=8 (`/usr/bin/time -l`), i.e.
  `tile_columns x plane_bytes` = 4 x 24.0 MB. t=1 flat at 99.5 -> 99.6 MB.
  Buffers are cached across frames and **not released at frame exit**, which is
  the strongest argument against defaulting this on as it stands.
* An earlier draft using `try_reserve_exact` + `resize(n, 0)` measured
  **+192.5 MB** — `Vec::resize` memsets, faulting in every page of an
  allocation whose point is that most of it is never touched. `vec![0u8; n]`
  lowers to `alloc_zeroed`. Comment at the site says so.
* **The carrier is free**: `tkoff` vs `base` overlaps on all 8 cells, unlike
  #473's tile key (+1.6-2.1% at 10bpc, from widening `WithOffset`).

## Two things for the index

* **`v4k_8tile` is 4:4:4, not 4:2:0** — all three planes 8,355,840 B, stride
  3840, `uv_stride == y_stride`. #473's `tile_stitch_cost.rs` priced the stitch
  against "3840x2160 4:2:0, 12.44 MB/frame"; the real frame is 25.1 MB, so that
  0.315 ms/frame figure is for a frame half this size.
* **`md5_inventory` gained `--threads`, and at t>1 `main`'s DEFAULT build
  panics** on the first film-grain vector:
  `thread_task.rs:534:56 called Option::unwrap() on a None value`
  (`ttd.delayed_fg.try_write().unwrap()`). Reproduced at t=2 and t=8 with no
  tile-owned code compiled in — **pre-existing**, and the reason the threaded
  corpus rows below cover 755 of 768. The corpus gate had never been run
  threaded; #455's verification round predicted exactly this blindness.

## Gates

* Corpus **766/766**, set-diffed BY NAME with the md5 as the value, against the
  committed baseline: **0 differing**. Threaded base-vs-head at t=1, t=2, t=8
  and head-t1-vs-head-t8: **0 differing** each (755 entries, film-grain
  excluded per above).
* Frame md5 identical to base at 8bpc t=1/2/4/8 and 10bpc t=1/2/4/8/16.
* `multi_decoder_pressure.sh` 12 decoders x 3 iters x {1,2,4,8,16} on 5 vectors:
  **PASS**. `decode_md5_verify` 14/14. lib 77/77.
* **6 new mechanism tests**, teeth proved by **3 mutations** (predicate ignoring
  `private`; stitch dropping the last row; final short sbrow rounded up instead
  of clamped), each restored by sha256 + `git diff --exit-code`.
* **The decode-md5 gate's own teeth measured at both ends**: the "drop the last
  row" mutation moves the t=8 frame md5 (`a00c11f4…` -> `0bcfec0b…`) and does
  NOT move t=1. So an identical md5 at t>1 is a real result and a t=1-only
  corpus run gates nothing here.
* Both standing hazards re-planted under `--features __probe_wide` (4af62ae
  in-lock `state` re-read deleted; `active()` cut to one shard):
  `wide_exclusion` **FAILS 3/3** on each, restored byte-exact.
* `#![forbid(unsafe_code)]` proved ACTIVELY — planted `unsafe` gives
  `error: usage of an unsafe block`; restored. `cargo fmt --check` clean.

## Not done

x86_64, Miri, `--features unchecked`, t=16 in the timed grid, any vector with
loop restoration live (#455 item 4 unchanged). Frame threading, intrabc frames
and c-ffi allocators all decline to the shared picture and were not exercised
through this path. Composing with #469. The tail of `cargo test` after
`decode_permutations` — the niced run was stopped to free the box for the timed
sweep and is re-running.


lilith:
### Follow-up on the same branch: the per-site census, and the completeness question answered

`--features tile-owned-recon,probe-sites`, v4k_8tile 8bpc t=8, `lost=0` on both
arms (`benchmarks/tile_owned_recon_sites_{off,on}_2026-08-09.txt`). The task
asked what fraction of recon writes actually went through tile buffers, because
a conversion that misses sites leaves the tracker on the path. **All of them:**

| site | OFF | ON | |
|---|---|---|---|
| `picture.rs:2061` `compact_read_per_row` | 2,662,876 | **0** | the split copy, gone |
| `picture.rs:168` `with_pixel_guard_immut` per-row | 2,336,708 | **0** | gone |
| `safe_simd/itx_arm.rs:8725` | 1,864,816 @ 8.0 B | **233,102 @ 26,888 B** | h per-row -> ONE hull |
| `safe_simd/ipred_arm.rs:1515` | 1,232,696 | **0** | gone |
| `ipred.rs:532` | 736,520 | **0** | gone |
| `safe_simd/itx_arm.rs:8894` | 564,304 | **0** | gone |
| `ipred.rs:1238` | 299,968 | **0** | gone |
| `picture.rs:173` `narrow_guard` hull | 0 | **280,464 @ 28,154 B** | the private branch, live |
| `loopfilter.rs:566` | 3,835,042 | 3,835,042 | filter chain, untouched by design |
| `cdef_arm.rs` + `cdef_apply.rs` | 1,563,248 | 1,563,248 | filter chain |
| `ctx.rs:99` | 2,534,988 | 2,534,988 | a context array, never a picture |

The mean-bytes column is the proof, not the count: 8.0 B x 1.86 M becomes
26,888 B x 233 k — the same pixels reserved once per block instead of once per
row. No site is left holding the tracker on the reconstruction path; the entire
residual is the filter chain plus `ctx.rs`.

### Remaining test binaries, run individually

All with the feature on, all `0 failed`: `mt_stress` 1/1 (1/2/4/8/16 x 5) —
and `photo_4k` was verified to ENGAGE the feature (22,530,490 -> 11,944,742
adds/frame at t=8), so it is a live gate rather than a vacuous one;
`tile_threading_parity` 3/3, `thread_cleanup_test` 2/2, `panic_safety_test`
4/4, `managed_api_test` 4/4, `safe_simd_crashes` 15/15, `integration_decode`
1/1, plus the pre-existing ignores in `tile_threading_overlap` (3) and
`reproduce_overlap{,_scalar}` (7).

### One more index-worthy fact

**`cargo clippy --all-targets -- -D warnings` is already RED on `main @ ee07b00`
(79 errors)**, so neither arm can claim "clippy passes". What was checked
instead: **zero findings land in any file this branch touches**. The first draft
did add one (`drop` of a `Copy` value in a new test) — fixed, along with the
unused binding it left behind.

lilith:
## Closing: #482 built this design soundly and shipped it, and the memory blocker is closed

Triage of the 15 open PRs, 2026-08-13. This round said **"Not for merge"** in its own body; it was a
prototype that priced a design, and the design went on to ship in a different geometry.

`main` cites this PR by number in four places and treats it as the predecessor it is:

> `src/owned_recon.rs`: "PR #474 gave each tile a private full-plane `Rav1dPictureDataComponent` but
> kept the existing borrow primitives on it, so the tracker still ran on every reconstruction access
> — just uncontended."

| this round measured | what `main` says now |
|---|---|
| 8bpc t=8 1.876x → **1.599x**, free at t=1 | superseded — #481 priced the completed conversion and #482 landed ON that ceiling (`MUT_RECON_KERNELS.md:151`) |
| **peak RSS 106.3 → 202.4 MB** (+96.3 at 8bpc, +191.0 at 10bpc), "the strongest argument against defaulting this on as it stands" | `MUT_RECON_KERNELS.md:64` and `OWNERSHIP_MODELS.md:84`, quoted as the cost #482 had to remove — **and it did: +1.6 MB at 8bpc / +3.2 MB at 10bpc, 59x and 60x smaller** |
| `Vec::resize` memsets and faults every page; `vec![0u8; n]` lowers to `alloc_zeroed` (an earlier draft measured +192.5 MB) | the comment survives at the site |
| **`v4k_8tile` is 4:4:4, not 4:2:0** — three 8,355,840 B planes, so #473's stitch figure was for a frame half this size | correction accepted and carried |
| **`md5_inventory --threads` at t>1 panics on `main`'s DEFAULT build** on the first film-grain vector (`thread_task.rs:534`, `Option::unwrap()` on `None`) | filed as **#479 and FIXED** — `AGENT_BRIEF.md` §7 now says `--skip-group film_grain` is no longer needed and `--threads 8` reaches 766/766 with no filter |

That last row is the most valuable thing this round produced and it was a side effect: the corpus
gate had **never been run threaded**, and running it immediately found a live bug on unmodified
`main`.

The design's residual blocker — that a tile task reaches its buffers through `fc.data.try_read()`,
a **shared** reference, so `recon_planes` can only hand back interior mutability — is recorded at
`OWNERSHIP_MODELS.md` §1 and was solved by moving the band to `Rav1dTaskContext`, which is already
`&mut` in every recon signature.

### What is not on `main`

`docs/TILE_OWNED_RECON.md` and `benchmarks/tile_owned_recon_*_2026-08-09.*` (6 files). They describe
a full-plane-per-tile geometry that `main` deliberately does not have, and every number in them is
superseded by #482's own measurements of the shipped geometry. Reachable at `pull/474/head` =
**`df9c4dee41f3e900bcba881ec574728a9ccd91a8`**; indexed in `benchmarks/pr_triage_2026-08-13.meta`.

PR #475, opened 2026-08-09
compose #472 + #474: the populations are NESTED — 1.9% more registrations, 6.7-9.7% SLOWER, and the memory fix the brief asked for doubles peak RSS

lilith:
Composes #472 (strided-rectangle record, carrying #469 + #470) with #474 (owned per-tile recon) on `main` @ `ee07b00`, fixes #474's memory lifetime, and audits the gate family the campaign has been quoting. **Not proposed for turning the feature on** — the composition is a measured regression. Full record: [`docs/COMPOSE_RECT_TILEOWN.md`](../blob/perf/compose-rect-tileown/docs/COMPOSE_RECT_TILEOWN.md) + `benchmarks/compose_rect_tileown_*_2026-08-09.*` (7 files).

## The thesis was refuted, twice over

#474 predicted the two shrink **disjoint** registration populations, so they should add. They are **nested**.

Registrations/frame, `--features probe-sites`, `lost=0` everywhere:

| arm | 8bpc t=1 | 8bpc t=8 | 10bpc t=1 | 10bpc t=8 |
|---|---|---|---|---|
| `base` (main) | 7,924,706 | 22,700,725 | 10,853,960 | 21,119,397 |
| `rect` (#472) | 7,924,706 | 8,250,328 | 10,853,960 | 10,852,772 |
| `tko` (#474) | 7,924,706 | 13,372,343 | 10,853,960 | 16,665,593 |
| **`both`** | 7,924,706 | **7,975,358** | 10,853,960 | **10,904,612** |

14,776,019 registrations are removable above the t=1 floor at 8bpc. `rect` removes 14,450,397, `tko` removes 9,328,382 — the sum exceeds the population. Composed removes 14,725,367: **274,970 more than `rect` alone, 1.9%**. At 10bpc the composition is **51,840 worse** than `rect` alone.

And it is slower. `n=9`, two-point wall fit (2 and 20 frames), rotating arm order, dav1d 1.5.4 `--framedelay 1` in the same sweep, `foreign_max = 0` on all 648 committed rows:

| head/base | 8bpc t=1 | t=2 | t=4 | t=8 | 10bpc t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|---|---|---|---|
| `both`/`bothoff` — **one binary**, runtime switch | 0.994 ᵒ | **1.067** | **1.091** | **1.091** | 0.998 ᵒ | **1.069** | **1.076** | **1.097** |
| `both`/`rect` — two builds | 0.987 ᵒ | **1.055** | **1.080** | **1.097** | 1.003 ᵒ | **1.062** | **1.082** | 1.092 ᵒ |

Bold = bands disjoint. ᵒ = OVERLAP. t=1 is null by construction (the feature declines below two workers).

**Why, given it deletes 326,810 compact copy pairs per frame.** Per-site diff `rect`→`both` shows the copies are *relocated*, not removed: `picture.rs:2159` (compact read/write-back) 326,810 → **0**, and `tile_recon.rs:358` (per-sbrow stitch) 0 → 25,920 borrows that move every reconstructed pixel of the frame once more. Traffic conserved; working set `tile_columns ×` larger.

## Gap to dav1d — best arm is #472 alone

ms/frame, median of 9. 8bpc: base 316.7 / 197.9 / 106.8 / 68.1 · **rect 325.3 / 177.5 / 92.9 / 56.9** · tko 320.8 / 185.6 / 96.4 / 59.2 · both 321.2 / 187.3 / 100.3 / 62.4 · dav1d 246.4 / 125.2 / 65.6 / 36.1.

ratio: base 1.285 / 1.581 / 1.627 / 1.885 · **rect 1.320 / 1.418 / 1.417 / 1.575** · tko 1.302 / 1.482 / 1.470 / 1.638 · both 1.303 / 1.497 / 1.529 / 1.728. Scaling t1→t8: base 4.65× · both 5.15× · tko 5.42× · **rect 5.72×** · dav1d 6.82×.

**~1.3× is met at no cell this round improves.** The only cell at or under the bar is 8bpc t=1, where **`base` is 1.285 and every arm here makes it slightly worse**. 10bpc best is rect at 1.440 / 1.503 / 1.497 / 1.702.

## Memory — the briefed fix was measured and rejected

`/usr/bin/time -l`, 20 decodes, n=3, `v4k_8tile` t=8:

| | 8bpc (base 111.4 MB) | 10bpc (base 164.9 MB) |
|---|---|---|
| feature off | +0.0 | +0.0 |
| **ceiling declines** (`MAX_MB=64`) | **+0.0** | **+0.0** |
| reuse, no flush release (= #474) | +100.8 | +200.3 |
| **reuse + release on flush (shipped)** | **+201.5** | **+402.4** |
| release at every frame exit | +205.7 | +402.4 |

1. **Releasing at frame exit doubles peak RSS.** macOS does not return a freed 25 MB region before the next frame faults a fresh set in — two generations resident. Kept as `RAV1D_TILE_OWNED_RELEASE=1`.
2. **Releasing on flush costs the same doubling** whenever decoding resumes after the flush. Measured, not inferred, via `RAV1D_TILE_OWNED_KEEP_ON_FLUSH=1`: +100.8 keeping vs +201.5 releasing. On this platform there is no cheap free-and-retake; it is a genuine trade — retention when idle, or 2× peak when decoding resumes.
3. **Shipped**: reuse across frames + `release_all()` from `rav1d_flush` (an idle decoder holds *nothing*, which was the actual complaint) + a ceiling `RAV1D_TILE_OWNED_MAX_MB`, default 256, charged on the allocation before it happens.
4. **Peak cannot go near baseline in this design.** Resident is `tile_columns × plane_set_bytes` — a tile writes a narrow column band of every row it owns and a 16 KiB page spans ~4 rows of a 3840-byte stride, so the tile's whole row range goes resident. Allocating only the tile's own rows does not help (`alloc_zeroed` never faults the phantom prefix) and no affine offset can change a stride. Only a column-**compact** tile buffer fixes it — #473's 22-site coordinate blocker. Not attempted.

### One arm was measuring nothing, and a counter caught it

The first 10bpc sweep's `both` arm **silently declined** (398 MB of 10-bit planes over the new 256 MB ceiling) and reported a clean null. Caught because its registration count came back **byte-identical to `rect`'s** (10,852,772); with `MAX_MB=1024` it is 10,904,612. All 10bpc cells were re-measured with the ceiling raised, and only then did 10bpc show the same regression as 8bpc.

## Dead-gate audit (the brief asked; here is what is there)

1. **`tests/decode_md5_verify.rs` — the documented 766/766 corpus gate — runs entirely at `threads = 1`** (`Settings::default()`, `src/managed.rs:196`). Every "766/766" in this campaign, including ones offered as cover for tile-threading changes, is a single-threaded number. Same defect as `md5_inventory`, in the test CI runs. **Not fixed here** — it needs a caller-visible threads knob plus a CI leg and should land on its own.
2. **`tests/decode_permutations.rs` also runs at `threads = 1`.** Defensible for its purpose, but the SIMD-permutation × tile-threading cross product is untested — and that is exactly where the `macos-15-intel` `compact_read_per_row` panic lives.
3. **`tests/tile_threading_parity.rs` does have teeth** — over **three committed vectors**, against 766 single-threaded ones.
4. **A fourth, already-known-silent one bit again**: the wasm32 cross job runs `cargo check`, which stops before the MIR pass raising `arithmetic_overflow`, so a test-side `1usize << 32` in `tracker_shard.rs` (a hard error where `usize` is 32 bits, and the i686 legs run `cargo nextest --lib` over both workspace members) was invisible. cfg-gated here.

### Running the corpus threaded immediately found a live bug on `main`

`av1-1-b8-23-film_grain-50` at `--threads 8`: `overlapping DisjointMut: current: &mut _[0..36864] … existing: &mut _[0..36864]` — **the same whole plane, twice**. `rav1d_apply_grain_row` hands each worker one row band, but `src/safe_simd/filmgrain_arm.rs:1550` takes `dst_row.full_guard_mut::<BD>()` over the entire component; the dead worker then wedges `thread_task.rs:534`. **Reproduced on unmodified `ee07b00`** with none of this branch compiled in. Not fixed here. It caps the threaded corpus at 753/766, which is why `md5_inventory` grew `--skip-group` — the exclusion belongs in the invocation where a set-diff can see the run shrink.

## Gates

- Corpus, set-diff **by name with the md5 as the value**: t=1 **766/766**. Four threaded legs (`base`/`both` × t=1/t=8): **755 rows each, 0 differing**.
- Frame md5 identical across **40 cells** (2 depths × t∈{1,2,4,8,16} × 4 arms) — exactly one hash per vector.
- `cargo test --release --features tile-owned-recon`: **164 passed, 0 failed** (incl. `decode_permutations` 19/19, `mt_stress` 1/2/4/8/16).
- `multi_decoder_pressure.sh` 12 × 3 × {1,2,4,8,16}: **PASS**.
- Miri `rect_hull_aliasing`, all **9 tests in isolation**: 9/9.
- `rect_oracle` at `RECT_ORACLE_ITERS=200000`: 2/2 (1.79 s vs 0.02 s at 2 k, so the env var really applied).
- Both standing hazards re-planted under `--features __probe_wide` → `wide_exclusion` **FAILS** each; restored byte-exact (`shasum -c` + empty `git diff`).
- `forbid(unsafe_code)` proved **actively** (planted `unsafe {}` → `usage of an 'unsafe' block`). No `unsafe` added.
- New `tests/tile_recon_lifetime.rs` with **teeth by two mutations**, each restored sha256-exact.

## Not measured

x86_64 (do not extrapolate — the panicking loop-filter compact path is `#[cfg(target_arch = "x86_64")]`); t=16 and oversubscription; anything below 4K in the timed grid; any vector with loop restoration live (unchanged campaign-wide blindness); negative strides; `--features unchecked`; film grain at t>1.

`tile-owned-recon` remains a non-default cargo feature, off at compile time.


lilith:
### Index correction: the campaign's best cells (1.306 / 1.328 / 1.333 / 1.470) belong to #469, not to a sound tree

Worth pinning, because those four numbers keep getting quoted as "where the rectangle got us". They were measured on **#469**, whose guard hands out `&mut rect.hull()` — the aliasing #470 found under Miri and #472 fixed. #472 keeps the single registration but hands out one ROW at a time, and its own report says the win drops to about two thirds.

Measured here on an idle box, `foreign_max = 0` on all 648 rows, n=9, dav1d 1.5.4 interleaved:

| 8bpc, ratio to dav1d | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| base (main @ ee07b00) | 1.285 | 1.581 | 1.627 | 1.885 |
| #469 as verified (quoted, unsound) | 1.306 | 1.328 | 1.333 | 1.470 |
| **#472 (sound), this sweep** | **1.320** | **1.418** | **1.417** | **1.575** |

My `base` column reproduces the verification round's (1.283 / 1.583 / 1.623 / 1.903) to within 1 %, so the arms are comparable. And my `rect` reproduces #472's own load-tagged head/base at t=2 almost exactly (its 0.896 × my base = 1.417 against my 1.418); at t=4 mine is worse and at t=8 better, which is what a load-tagged column against an idle one looks like.

**So the sound rectangle costs roughly a third of the headline.** The campaign index's "closest this has been at any thread count above 1" cells are not reachable from a tree without the aliasing.

### The 32-bit break, A/B'd rather than argued

`cargo build --target wasm32-wasip1 -p rav1d-disjoint-mut --features std --all-targets` — clean at this HEAD; with the new `#[cfg(target_pointer_width = "64")]` removed it fails with `this arithmetic operation will overflow` / `#[deny(arithmetic_overflow)]` in the **lib test** target. `cargo check` with identical arguments is green in **both** states. Third time this campaign that `check` has been read as a build gate.

lilith:
### CI, leg by leg

**`CI` workflow: 21/21 green**, including every leg the standing rules care about.

| leg | result |
|---|---|
| Build & Test — ubuntu safe-simd / asm / **i686** | success ×3 |
| Build & Test — windows safe-simd / **i686** / **windows-11-arm** | success ×3 |
| Build & Test — macos-latest / **macos-15-intel** | success ×2 |
| Build & Test — ubuntu-24.04-arm safe-simd-arm64 | success |
| Conformance (decode permutations) aarch64 / x86_64 | success ×2 |
| SIMD Permutation Tests aarch64 / x86_64 | success ×2 |
| Tile Threading · Threading race gates | success ×2 |
| Cross-compile musl / aarch64-gnu / wasm32-wasip1 | success ×3 |
| Clippy · Format · Coverage | success ×3 |

Two worth calling out: **both i686 legs are green**, which is what the `1usize << 32` cfg gate was for — they run `cargo nextest --lib` over both workspace members and would have hit it. And **`macos-15-intel` is green this run** — the known intermittent `compact_read_per_row` panic (x86_64-only site, also fails on plain `main`, run 31253865532) did not fire. One green run is not evidence it is gone.

**`disjoint-mut CI`: 3 legs recovered, 1 still red, Miri still running.**

| leg | on `main` @ ee07b00 | here |
|---|---|---|
| `Test (ubuntu/macos/windows, --no-default-features)` | **failure ×3** | **success ×3** ← fixed |
| `Test (ubuntu-latest, )` ×3 platforms | success | success |
| `Wide-path exclusion gate (non-vacuous)` · Lint · Docs · MSRV | success | success |
| `Test (ubuntu-latest, --all-features)` | failure | failure (pre-existing) |
| `Miri (Object)` ×2 | **failure + cancelled** | in progress |

The `--all-features` leg asks for a self-contradictory configuration: `__probe_count` / `__probe_noscan` / `__probe_lockonly` / `__tracker_legacy` each *replace* the sharded tracker, so turning them all on at once selects an untracked one while `test_new_always_tracked` asserts `v.is_checked()`. This PR fixes the part that was a plain compile break (`wide_exclusion`'s probe block spelled only `__probe_wide` where the lib's re-export of `wide_probe` also requires the sharded tracker, so the binary would not build and took the rest of the job with it). The remaining two lib-test failures want either the leg dropped or a `compile_error!` making the selectors mutually exclusive — a decision for the crate, not something to paper over from here.

**`Miri (Object)` is red on `main` for a reason that deserves its own issue**, and it is not this PR's: run 31300... on `ee07b00` finished after **1 h 46 m** with

```
error: Undefined Behavior: not granting access to tag <126804> because that would
       remove [Unique for <2134660>] which is strongly protected
error: test failed, to rerun pass `-p rav1d-disjoint-mut --test narrow_release`
```

after `26 passed; 0 failed` in the preceding binary. This branch's Miri jobs are still running at the time of writing; the locally-run Miri gate that matters for the rectangle work — `rect_hull_aliasing`, all **9 tests each in its own process** — is 9/9 clean.

lilith:
### Miri landed — same pre-existing failure as `main`, same test, same shape

`disjoint-mut CI` / `Miri (Object)` finished after **2 h 00 m** (11:01:58 → 13:02:00 UTC). Here:

```
test result: ok. 35 passed; 0 failed; …  finished in 1521.25s
error: Undefined Behavior: not granting access to tag <127146> because that would
       remove [Unique for <7735178>] which is strongly protected
error: test failed, to rerun pass `-p rav1d-disjoint-mut --test narrow_release`
```

`main` @ `ee07b00`, run `31280…`:

```
test result: ok. 26 passed; 0 failed; …  finished in 1489.21s
error: Undefined Behavior: not granting access to tag <126804> because that would
       remove [Unique for <2134660>] which is strongly protected
error: test failed, to rerun pass `-p rav1d-disjoint-mut --test narrow_release`
```

Same test, same Stacked-Borrows shape, only the tag numbers differ. **Pre-existing; not introduced here.** The preceding binary went 26 → 35 passing because #470/#472 add Miri cases, which is also why the job ran ~14 minutes longer.

The Miri gate that matters for the rectangle work is clean: `rect_hull_aliasing`, all **9 tests each in its own process** (Miri aborts on first UB, so a batched run proves less), 9/9.

**Final tally for this PR** — `CI` **21/21 green**; `disjoint-mut CI` 10 green (3 of them recovered from red-on-main), `Test (ubuntu-latest, --all-features)` red (self-contradictory feature set, pre-existing), `Miri (Object)` red on `narrow_release` (pre-existing, reproduced on `main`). No leg is red because of this branch.

lilith:
## Closing: composing two superseded prototypes — and its best finding, the dead-gate audit, is FIXED on `main`

Triage of the 15 open PRs, 2026-08-13. This round composed #472 with #474 and measured the
composition as a **regression** ("Not proposed for turning the feature on"), so there was never
anything to merge. Both parents have since been superseded — #472's row view by #505/#513, #474's
owned recon by #482's column-compact band — which makes the composition moot as well as negative.

### The findings, and where they are

* **"the populations are NESTED, not disjoint"** — 14,776,019 removable above the t=1 floor at 8bpc;
  `rect` removes 14,450,397, `tko` removes 9,328,382, the sum exceeds the population, and composed
  removes only 274,970 more than `rect` alone. That is a real result and it stands. `main` reached
  the same conclusion structurally: #482 removes recon's split and #505 collapses the filter chain's,
  and `MUT_RECON_KERNELS.md:532` records recon's 7.9M → 22.7M at t=8 as the split #482 removed.
* **the hull over-reserves inter-row gaps** — `BOUNDS_MAP.md:16` marks this arm **OVER-RESERVED
  336x-1680x, all inter-row gap**, and `bounds_probe.rs:5` names it in the instrument's own header
  as one of the "count reduction" attempts it exists to have predicted.
* **releasing at frame exit doubles peak RSS** on macOS — the structural point (residency is
  `tile_columns × plane_set_bytes`, and only a **column-compact** buffer fixes it) is `MUT_RECON_KERNELS.md`
  §"Geometry: column-compact, one superblock row, its own stride", which is what #482 built.
* **an arm that silently declined and reported a clean null**, caught only because its registration
  count came back byte-identical — the same "assert on the mechanism, not the outcome" lesson #473
  hit from a different direction.

### The dead-gate audit is the best thing here, and item 1 is fixed

> "`tests/decode_md5_verify.rs` — the documented 766/766 corpus gate — runs entirely at `threads = 1`.
> Every '766/766' in this campaign, including ones offered as cover for tile-threading changes, is a
> single-threaded number. **Not fixed here** — it needs a caller-visible threads knob plus a CI leg
> and should land on its own."

It landed on its own. `tests/decode_md5_verify.rs` now has `md5_threads()` reading
`RAV1D_MD5_THREADS`, with a hard error on a malformed value rather than a silent downgrade, and
`ci.yml` runs the corpus **twice** — once unset, once at `RAV1D_MD5_THREADS: 8`. The function's own
doc comment records why, and names what it would have caught: the x86_64 t=8 deblock race
(#494/#497).

**And running the corpus threaded here immediately found a live bug on unmodified `ee07b00`** —
`rav1d_apply_grain_row` vs `filmgrain_arm.rs:1550` taking a whole-component guard. Filed as #479 and
**fixed**; `AGENT_BRIEF.md` §7 now says the film-grain skip is no longer needed.

### What is not on `main`

`docs/COMPOSE_RECT_TILEOWN.md` and `benchmarks/compose_rect_tileown_*_2026-08-09.*` (7 files) — a
composition of two code paths neither of which exists on `main` in the measured form. Reachable at
`pull/475/head` = **`639db7de918c6417a12736eb178fb30d6f02ca68`**; indexed in

Issue #479, opened 2026-08-09
film grain takes a WHOLE-PLANE mutable guard per row worker: overlapping DisjointMut aborts every --threads>1 decode with apply_grain


lilith:
Pre-existing on `main` @ b0f70ee. Found while running the corpus gate at
`--threads 8` for #477; **not** caused by that PR — verified identical on a
pristine `main` worktree.

## Repro

```
cargo build --release --example md5_inventory     # after #478, which adds --threads
./target/release/examples/md5_inventory --threads 8 --group 8-bit/film_grain/meson.build
```

Aborts at the first vector (`av1-1-b8-23-film_grain-50`), rc=101:

```
thread 'rav1d-worker-5' panicked at src/safe_simd/filmgrain_arm.rs:1550:53:
	overlapping DisjointMut:
 current: &mut _[0..147456] on ThreadId(6) at src/safe_simd/filmgrain_arm.rs:1550:53
existing: &mut _[0..147456]                at src/safe_simd/filmgrain_arm.rs:1550:53
...
thread 'main' panicked at src/thread_task.rs:534:56:
called `Option::unwrap()` on a `None` value
```

(Call sites via `--features probe-sites`; without it both sides report the
`picture.rs:736` wrapper.)

## Cause

`src/safe_simd/filmgrain_arm.rs:1550` (and 1572, 1644, 1675) do

```rust
let (mut dst_guard, dst_base) = dst_row.full_guard_mut::<BD>();
```

`full_guard_mut` (`include/dav1d/picture.rs:1286`) registers
`slice_mut::<BD,_>((0.., ..total_pixels))` — **the entire plane**. Film grain runs
one task per `FG_BLOCK_SIZE` row band, so every worker registers the whole plane
mutably at the same time.

The *rows written* are disjoint. The *borrows registered* are not, and at the Rust
level N concurrent `&mut [u8]` over one plane is UB whatever the writes do. So the
tracker is right to refuse; the kernel is asking for the wrong thing.

`src/looprestoration.rs:212,258` and `src/loopfilter.rs:140,182` use the same
`full_guard_mut` shape and are worth checking for the same pattern.

## Impact

- Any `apply_grain` decode at `threads > 1` panics its workers and then the main
  thread. 13 of 768 corpus vectors (all of `8-bit/film_grain` and
  `10-bit/film_grain`) cannot be decoded at all at `--threads 8`.
- The corpus gate never noticed because `md5_inventory` and
[...]

PR #480, opened 2026-08-09
REFUTED: declining the strided rectangle at t=1 is a measured null — #472's t=1 cost is not the rectangle path

lilith:
**Do not merge.** This is the record for a hypothesis that measured NULL, kept for its gates and its ruled-out list.

## The hypothesis, and why it looked near-certain

#472 (the sound strided rectangle) wins big at t>1 and regresses 8bpc t=1, losing the campaign's only cell under the ~1.30x bar. Its own analysis says t=1 pays **+0.71 ns on every surviving registration while removing ZERO of them**, because the per-row split the rectangle collapses only exists once `tile_threading_active()` latches. So the rectangle should decline at t=1 and keep `main`'s path exactly.

`rect_is_exact` now carries the latch, taking all four rectangle construction sites — `for_rows`, `for_rows_mut`, `LfBlock::fill_hull`, and the two `compact_*_per_row` helpers — back to the plain strided hull when no tile worker can be alive. Proved dead by a `#[cfg(test)]` branch counter, not assumed.

## It buys nothing

n=9, 288 rows, `foreign_max = 0` on **all** of them, idle M4 Pro, dav1d 1.5.4 `--framedelay 1` in the same interleaved sweep, two-point wall fit, rotating arm order, 3 cell-rounds discarded and re-run. Paired within-round ratios (the arms in one round saw the same machine state), exact two-sided sign test:

| | median | band | p | |
|---|---|---|---|---|
| gated / #472, 8bpc t=1 | **1.0005** | [0.9916..1.0151] | 1.000 | NULL |
| gated / #472, 8bpc t=2/4/8 | 0.9972 / 1.0036 / 1.0050 | | 0.508 / 0.508 / 0.180 | NULL |
| gated / #472, 10bpc t=1 | **1.0109** | [1.0008..1.0157] | 0.004 | **WORSE**, 9/9 rounds agree |
| gated / main, 8bpc t=1 | **1.0175** | [1.0092..1.0372] | 0.004 | the cost SURVIVES |

Composite, ratio to dav1d, 8bpc t=1/2/4/8:

| | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| main (cebb97f) | 1.306 | 1.582 | 1.627 | 1.869 |
| #472 | 1.335 | 1.420 | 1.412 | 1.569 |
| **gated** | **1.331** | 1.420 | 1.415 | 1.577 |

The predicted "best of both" — 1.285 / 1.418 / 1.417 / 1.575 — **does not exist**. Full table with bands: `benchmarks/rect_t1_decline_2026-08-09.meta`, raw rows in `benchmarks/rect_t1_decline_gap_2026-08-09.tsv`.

## What that means

The +0.71 ns is paid by **every** registration, including the plain hull ones the rectangle never touches. Ruled out, each with its evidence:

* the rectangle **code path** — this round, paired 1.0005 at 8bpc t=1 with the path provably dead;
* the per-`alloc` **column stores** — #472's own `cols_ab` A/B (0.9966 / 1.0139 / 1.0037, every band overlapping, sign inconsistent);
* the **row map** being read on the hot path — by inspection, not timing: `add` dispatches `self.mask == 0` FIRST and returns before `self.rowmap.enabled()`, and every instance is `mask == 0` with no parallelism declared (`SHARDS_SERIAL == 1`).

Left, and **not** measured: `add`'s two **extra parameters**. #469 turned `add<IS_MUT>(&self, bounds)` into `add<IS_MUT>(&self, bounds, row_w: u32, rows: u32)` and `add_mut`/`add_immut` pass `0, 0` — two arguments materialised on 7,924,706 calls per frame at t=1, on a path whose own comments price the call barrier at 4% of the tracker and call it latency-bound. The obvious test (a 1-argument `add_flat` for the non-rect entry points) has a known trap: duplicating that body for a different reason already cost +0.9% on I-cache alone (`benchmarks/tracker_borrowcost_2026-08-08.tsv`, arm `fold`).

**Index correction:** `main` is **1.306** at 8bpc t=1 on cebb97f, not the 1.285 the ledger carries from ee07b00, and that cell's band [310.9..325.2] is the widest in this sweep (4.4%). t=1 sub-2% claims on this vector need the paired statistic, not the medians.

## The gates are worth keeping regardless

The hull-vs-rectangle choice is invisible **twice over**: identical pixels by construction, AND an identical tracker *record* — a serial big plane is single-shard, so `rect_exact_for` answers `false` and a rectangle registration degrades to exactly the hull it is being compared against. So the existing conflict-probe tests read the same on both branches and nothing in the tree can see a regression in either direction. Four `#[cfg(test)]` branch-counter tests, two per call site, each asserting the OTHER branch's counter is zero.

Teeth by mutation ×3, each restored byte-exact (sha256):

| plant | result |
|---|---|
| `for_rows_mut` un-gated | picture OFF half FAILS |
| rectangle deleted outright | picture ON half FAILS |
| `fill_hull` un-gated | loopfilter OFF half FAILS — and before that pair existed it left the **entire** `--lib` suite green (73 passed / 0 failed) |

Also here: `md5_inventory --skip-group` (repeatable, prints `skipped_groups=` on the TOTAL line) so the threaded corpus can run around #479 at all.

## Gates

* `cargo test --release --lib` 75 passed / 0 failed; `cargo fmt --all --check` clean.
* Corpus set-diffed **BY NAME with the md5 as the value** against `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`: `--threads 1` 766 PASS, 768 keys, **0 differing**; `--threads 8` 753 PASS, 754 keys after dropping the two film-grain groups from BOTH sides (#479), **0 differing**; and gated-t1 against a fresh pristine-`main` t1 inventory, **0 differing**.
* Frame md5 identical across main / #472 / gated at t = 1/2/4/8/**16**, both depths — 30 of 30.
* `mt_stress` 1/2/4/8/16 × 5. `multi_decoder_pressure` 12 × 3 × {1,2,4,8,16} PASS.
* Both standing hazards re-planted under `--features __probe_wide` — the 4af62ae in-lock `state` re-read deleted **from `add_contended` alone** (the contended path; the `try_lock` half is not gated) and `active()` cut to one shard — `wide_exclusion` FAILS each, restored byte-exact by sha256 **and** `git diff --exit-code`, green again after.
* `#![forbid(unsafe_code)]` proved ACTIVELY (planted `unsafe` → rejected at `lib.rs:13`).
* Miri on `rav1d-disjoint-mut` under both models: see the comment thread — running at the time of writing.

Not measured: x86_64; `--features unchecked`; any vector but the two 4K gap vectors, which are intra-only with loop restoration OFF while LR is live in 696 of 768 corpus vectors.


lilith:
### Miri status, both models — and the one target that did not finish

**Complete and green, `rav1d-disjoint-mut`, Apple M4 Pro, nightly `miri 0.1.0 (da86f4d072 2026-07-24)`:**

| target | Stacked Borrows | Tree Borrows |
|---|---|---|
| lib unittests | 35 ok | 35 ok |
| `aligned_miri` | ok | ok |
| `guard_move_release` | 2 ok | 2 ok |
| `narrow_release` | 1 ok | 1 ok |
| `pic_buf_overflow` | ok | ok |
| `rect_hull_aliasing` | 9 ok | 9 ok |
| `shard_liveness` | 5 ok (996 s) | 5 ok (996 s) |
| `soundness` | 25 ok | 25 ok |
| `wide_exclusion` | 1 ok | 1 ok |

**0 `error: Undefined Behavior` on either model.** `narrow_release` — the target that was the Stacked-Borrows UB report on `main` until #478 — is green here under both, so this composition does not regress what cebb97f just fixed.

**NOT finished, stated rather than glossed: `rect_oracle`.** Both `--no-fail-fast` legs are still inside `rect_versus_rect_and_rect_versus_interval_match_a_byte_set_oracle` after ~2 h each and were left running rather than reported as green. It is 1,600,000 randomized pairs over 8 plane shapes, and under Miri that is hours, not minutes. It is **#472's own test, unchanged by this branch** (this branch adds a latch term to `rect_is_exact` in `include/dav1d/picture.rs` and `src/loopfilter.rs`, and touches no code `rect_oracle` exercises), and #470 reported it Miri-clean when it landed. The other eight targets above were re-run individually to get complete coverage around it.

Suite: `cargo test --release` on this branch is **161 passed / 0 failed over 21 targets**.

lilith:
## Closing: its ruled-out list stands, its RESIDUAL ATTRIBUTION was refuted by #506

Triage of the 15 open PRs, 2026-08-13. This PR says **"Do not merge"** in its first line — it is a
record for a hypothesis that measured null — so closing it is what it asked for. But its conclusion
should not be carried forward as written, and that is the reason for a long comment on a
deliberately-unmergeable PR.

### What stands

The ruled-out list is good work and each item keeps its evidence:

* **the rectangle code path** — gated off, proved dead by a `#[cfg(test)]` branch counter rather
  than assumed, paired **1.0005** at 8bpc t=1, p=1.000;
* **the per-`alloc` column stores** — #472's `cols_ab` A/B, every band overlapping, sign inconsistent;
* **the row map on the hot path** — ruled out by inspection: `add` dispatches `self.mask == 0` first
  and returns before `self.rowmap.enabled()`.

Its index correction also stands and is the right instinct: `main` was **1.306** at 8bpc t=1 on
`cebb97f`, not the 1.285 the ledger carried from `ee07b00`, and that cell's band was the widest in
the sweep at 4.4%.

### What does not

The conclusion — *"The +0.71 ns is paid by **every** registration, including the plain hull ones the
rectangle never touches"* — and the one suspect it left open, `add`'s two extra parameters.

**#506 measured the answer and it is neither.** It is **CODE PLACEMENT**:

| arm | added `__text` | wall/plain at `v4k8tile` t=1 | sign |
|---|---|---|---|
| byte-identical copy | 0 | 1.0006 | 4/11 |
| provably-dead `#[used]` text, **0 symbols resized** | +4,828 | **1.0110** | **0/11** |
| … +9,692 / +14,556 / +19,420 | | 1.0117 / 1.0109 / 1.0156 | 0/11 each |

Nine binaries differing from `main`'s by +1,132 B to +19,420 B — including a refactor that only
*shrinks* the one changed function by 17% — all land in **+1.1% to +1.6%**. Against a same-source
control the rectangle costs **0.9967 (7/9)** at t=1. The tax scales with working set: +1.4% at 4K,
+0.7% at 1024x576, **0** at 256x2048.

So this round's premise — that a t=1 delta measured against `main`'s binary is attributable to
anything in the diff — does not hold at this magnitude, and the hypothesis it was testing could not
have been confirmed or refuted by that comparison in either direction. The gating experiment was
sound; the baseline was a lottery ticket. `main` carries the rule now (`AGENT_BRIEF.md` §6 and §2):
**judge a t=1 claim inside ~1.5% against a same-source arm or a `__pad_text` rung, never against
"the same source built in a second worktree"**, which moves nothing and reads ±0.1%. #506 also
measured the lottery to be mostly removable at **16-byte function alignment** (rung spread 1.37% →
0.18%), which is now a measurement flag rather than an open question.

### What is not on `main`

`benchmarks/rect_t1_decline_2026-08-09.meta` + its 288-row TSV. They are a null on a code path
`main` does not have, measured against a baseline the campaign has since learned not to trust at
that magnitude. Reachable at `pull/480/head` = **`ae0dd65f2d0932e993c21930ad0bfa9b8144020b`**;
this correction is recorded on `main` in `benchmarks/pr_triage_2026-08-13.meta` under
"Two stale conclusions on the closed branches".

Its gate design is worth remembering independently of the result: the hull-vs-rectangle choice is
invisible **twice over** — identical pixels AND an identical tracker record on a serial big plane —
so nothing in the tree could see a regression in either direction, and it built four branch-counter
tests plus three mutations to fix that. One of those mutations left the **entire** `--lib` suite
green before the matching test pair existed.

PR #481, opened 2026-08-09
Zero-tracker recon path, PRICED: ceiling is 1.276/1.333/1.313/1.475 at 8bpc — worth building, and the blocker is +191 MB

lilith:
**Do not merge.** Measurement round. Every behavioural change is behind a `probe-`/`__`-gated feature, absent from `default` and from every published feature.

Companion doc: [`docs/ZERO_TRACKER_RECON.md`](https://github.com/imazen/rav1d-safe/blob/perf/zero-tracker-recon/docs/ZERO_TRACKER_RECON.md). Base `main` @ `cebb97f`, composed with #474.

## The question

#474 built owned per-tile recon buffers but **kept the existing borrow primitives on a private instance**, so the tracker is still on the recon path, at its uncontended t=1 shape. Its author named the remaining step: a `&mut [u8]` refactor of every kernel signature taking `WithOffset<&Rav1dPictureDataComponent>` — ~27 call sites plus ~40 `with_pixel_guard_mut` closures.

A **partial** conversion is uninterpretable: it leaves the tracker on the path, the win does not appear, and that reads as "the design fails" when it means "the conversion is incomplete". So this prices the **completed** conversion instead of writing part of one.

## The instrument

`--features probe-recon-untracked` builds the private per-tile planes with `tracker: None` and leaves the shared picture — and the whole filter chain on it — tracked exactly as today. By construction that is the union of what a complete conversion can remove.

`#![forbid(unsafe_code)]` is **not** relaxed in `rav1d-safe`: `forbid` cannot be locally allowed, so the sub-crate (which already holds the `unsafe` the tracker encapsulates) exposes a safe `DisjointMut::probe_untracked` behind `__probe_untracked_ctor`, and the consumer calls a safe function.

**Liveness by COUNT and BY NAME, not by timing.** `--features probe-sites`, `v4k_8tile` 8bpc, `lost=0`:

| arm | t=1 | t=8 | sites |
|---|---|---|---|
| base | 7,924,706 | 22,700,725 | 68 |
| tko (#474) | 7,976,546 | 13,372,343 | 70 / 69 |
| **ceil** | **6,005,602** | **11,401,399** | 49 / 48 |

`base` and `tko` reproduce #474's committed census **bit for bit**. Exactly **1,970,944** registrations vanish, the **same number at t=1 and t=8**, across 21 sites — every one recon (`ipred*`, `itx_arm`, `ipred_prepare`, the stitch's read side, `with_pixel_guard_immut`), **not one** from the filter chain or the `BlockContext` class.

## The price, and it is not small

n=9, 288 rows, `foreign_max = 0` on **all** of them, idle M4 Pro, dav1d 1.5.4 `--framedelay 1` in the same interleaved sweep, two-point wall fit, rotating arm order, one cell-round discarded. `RAV1D_TILE_OWNED_T1=1` throughout (new knob — #474 declines below two workers, and the ceiling has to be priced at the cell the bar is closest at).

Ratio to dav1d, 8bpc:

| | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| base | 1.296 | 1.578 | 1.632 | 1.883 |
| tko (#474) | 1.299 | 1.481 | 1.478 | 1.628 |
| **zero-tracker recon** | **1.276** | **1.333** | **1.313** | **1.475** |
| #472 (sound rectangle, re-measured this session) | 1.335 | 1.420 | 1.412 | 1.569 |
| untracked ceiling (whole tracker gone) | 1.160 | 1.264 | 1.262 | 1.345 |

10bpc: 1.393 / 1.452 / 1.467 / 1.610 against base 1.432 / 1.617 / 1.662 / 1.926.

Paired within-round `ceil / tko`, exact sign test: **0.9806 / 0.8996 / 0.8955 / 0.9071** at 8bpc t=1/2/4/8, p=0.004, **bands disjoint at all four**. `ceil / base`: 0.9833 / 0.8437 / 0.8087 / 0.7708.

**It beats #472 at every one of the eight cells**, and 8bpc t=1 at 1.276 is the only arm this campaign has produced under the ~1.30x bar. t=2 and t=4 land within 0.07 and 0.05 of the whole-tracker ceiling. **t=8 still misses at 1.475** and no recon work closes it — the ceiling there is 1.345.

**My own registration-rate estimate said the opposite and was wrong.** 1.97 M removed bought 18.5 / 10.7 / 5.6 ms at t=2/4/8 — **9.4 / 5.4 / 2.8 ns per removed registration**, against the 1.00 ns the #469 verification round measured for the rectangle's removals. Count is not cost in **both** directions, and that constant does not transfer between site classes: this population's mean extent runs 10 B to 58 KB with a heavy tail of 26–32 KB strided hulls, and each carries guard-drop glue that goes with the record.

## The blocker is memory, and the fix is the same change

Peak RSS, `/usr/bin/time -l`, 20 frames:

| arm | 8bpc t=8 | 10bpc t=8 |
|---|---|---|
| base | 106.2 MB | 157.3 MB |
| tko | 202.5 | 348.3 |
| ceil | 202.0 | 347.8 |

**+96.3 MB at 8bpc, +191.0 MB at 10bpc.** `ceil` and `tko` differ by 0.5 MB, so it is not the tracker — it is structural: `v4k_8tile` is 4:4:4, three 8,355,840-byte planes, tiled 4×2, so each of 8 private full-plane buffers has half its rows touched.

**Deliberate design statement, since the round was asked for one: the shippable geometry is column-compact, and it is not a follow-up — it is the same change.** #474 bought its one-expression seam by giving each tile buffer the picture's byte length AND stride, which is exactly what makes the whole plane addressable and the residency `tile_columns × plane`. A kernel signature taking `(&mut [BD::Pixel], stride)` already carries the origin and stride a compact buffer needs. Doing the `&mut` conversion first and the geometry second pays #473's 22-site coordinate cost twice.

## What this round did NOT do

* **The refactor is not written.** Zero of ~27 call sites, zero of ~40 closures. Every number above is from an UNSOUND probe.
* **The `&mut` is not obtainable where #474 puts the buffers**, and this is an uncosted prerequisite: tile tasks reach `f.tile_recon` via `fc.data.try_read()` → `&Rav1dFrameData` (`src/thread_task.rs:1237`), a **shared** reference, so `recon_planes` can only hand back `&Rav1dPictureDataComponent` — interior mutability, i.e. the tracker. The buffers must first move to something the task owns exclusively; `t: &mut Rav1dTaskContext` is the obvious candidate and is already `&mut` in every recon signature. Same shape as #468's blocker, one level in — check it in a minute, don't discover it in an afternoon.
* **Composition with #472 was not re-tested.** #475 measured it negative (nested populations) and nothing here changes that.
* Not measured: x86_64; Miri on the probe arm; `--features unchecked`; t=16; anything below 4K; **any vector with loop restoration live** — `looprestoration*.rs` registers zero borrows in this grid while LR is active in 696/768 corpus vectors, so §3's filter share is understated and the recon share correspondingly overstated.

## Gates

* Corpus set-diffed **BY NAME with the md5 as the value**, both arms: `--threads 1` 766 PASS / 768 keys / **0 differing**; `--threads 8` 753 PASS / 754 keys after dropping the two film-grain groups from BOTH sides (#479) / **0 differing**. This is the first time the private-buffer path has been corpus-gated at **t=1** at all — #474 declines below two workers, so its t=1 run exercised none of it.
* `md5_inventory --skip-group` added (repeatable; prints `skipped_groups=` on the TOTAL line so an empty filter cannot pass as a full run).
* Frame md5 identical across base / tko / ceil at t=1/2/4/8, both depths — 24 of 24.
* `mt_stress` 1/2/4/8/16 × 5; `multi_decoder_pressure` 12 × 3 × {1,2,4,8,16} **PASS on tko AND ceil**.
* Both standing hazards re-planted under `--features __probe_wide` — the 4af62ae in-lock `state` re-read deleted from `add_contended` **alone**, and `active()` cut to one shard — `wide_exclusion` FAILS each; restored byte-exact by sha256 **and** `git diff --exit-code`.
* `#![forbid(unsafe_code)]` proved ACTIVELY — **and the first proof was vacuous**: an `unsafe` block planted in `src/tile_recon.rs` builds cleanly on the default feature set, because that whole module is `#[cfg(feature = "tile-owned-recon")]` and is not compiled. With the feature on it is rejected at `lib.rs:13`; a second plant in `include/dav1d/picture.rs` is rejected on the default build. **A forbid proof on a cfg'd-out file proves nothing.**
* Miri on `rav1d-disjoint-mut`: see the comment thread.


lilith:
### Miri: both models COMPLETE and green

`cargo +nightly miri test -p rav1d-disjoint-mut --no-fail-fast`, Apple M4 Pro, `miri 0.1.0 (da86f4d072 2026-07-24)`:

| model | result |
|---|---|
| Stacked Borrows (default) | **rc=0**, 9 targets, 61 tests, **0 `error: Undefined Behavior`** |
| Tree Borrows (`-Zmiri-tree-borrows`) | **rc=0**, 9 targets, 61 tests, **0 UB** |

Per-target both models: lib unittests 26, `guard_move_release` 2, `narrow_release` 1, `rect_hull_aliasing` 0 (absent on this base), `shard_liveness` 5 (993 s), `soundness` 25, `wide_exclusion` 1, `aligned_miri` 1, `pic_buf_overflow` 0. `narrow_release` — the Stacked-Borrows UB report that was red on `main` until #478 — is green under both, so this branch does not regress it.

That is the expected result and worth stating why it is weak evidence for the *headline*: the only change this branch makes to the sub-crate is the `__probe_untracked_ctor` constructor, which is **off** in every configuration Miri runs here. Miri is a gate on the composition not regressing, not a check on the probe arm — **the probe arm is UNSOUND by construction and was not Miri'd**, and it should not be, since it deletes exactly the bookkeeping Miri would be checking against.

lilith:
## Closing: this priced a change, and #482 WROTE it — `main` says so in its first five lines

Triage of the 15 open PRs, 2026-08-13. This PR opens **"Do not merge. Measurement round."** and it
did the job a pricing round exists to do: it made the case, and the case was built.

`docs/MUT_RECON_KERNELS.md` line 5 on `main`:

> **PR #481 priced this change and did not write it. This writes it**, on a band that is
> column-compact instead of full-plane, so the memory blocker #481 named is …

And `src/owned_recon.rs`'s module doc:

> PR #481 priced removing it outright with a `tracker: None` probe: **1,970,944 registrations per
> frame across 21 sites**, worth 0.8996 / 0.8955 / 0.9071 of #474's wall at 8bpc t=2/4/8. **This
> module earns those numbers SOUNDLY.**

Its unsound ceiling is the target `main` measured itself against — `MUT_RECON_KERNELS.md:151` has a
row "**#481's unsound ceiling** | *1.276* | *1.333* | *1.313* | *1.475*", and
`benchmarks/mut_recon_2026-08-09.meta:28` repeats it with the 10bpc column. Its
**2.8-9.4 ns/registration** band is cited three times (`CTX_TL_SPLIT.md:207`,
`MUT_RECON_KERNELS.md:662` and `:701`) as the range later measurements are checked against.

### The design statement it was asked for was accepted verbatim

> "the shippable geometry is **column-compact**, and it is not a follow-up — it is the same change …
> Doing the `&mut` conversion first and the geometry second pays #473's 22-site coordinate cost twice."

That is what #482 built: one superblock row tall, own stride, one band per WORKER. Memory went from
this round's **+96.3 / +191.0 MB** to **+1.6 / +3.2 MB** (`OWNERSHIP_MODELS.md:84`) — 59x and 60x.

The prerequisite this round flagged was also right and was solved the way it suggested: tile tasks
reach `f.tile_recon` through `fc.data.try_read()` → a **shared** reference, so the buffers had to
move to something the worker owns exclusively, and `t: &mut Rav1dTaskContext` was "the obvious
candidate" — which is where `owned_recon`'s band lives.

Two process lessons from this round that `main` now carries as rules:

* **a partial conversion is uninterpretable** — it leaves the tracker on the path, the win does not
  appear, and that reads as "the design fails" when it means "the conversion is incomplete". Hence
  pricing the completed conversion instead of writing part of one.
* **"a forbid proof on a cfg'd-out file proves nothing"** — the first `#![forbid(unsafe_code)]` proof
  here was vacuous because `src/tile_recon.rs` was entirely `#[cfg(feature = "tile-owned-recon")]`.
  `MUT_RECON_KERNELS.md` §6 is titled "proved actively and **NON-VACUOUSLY**" and opens by citing
  this. Self-disclosed rather than hidden, which is why it became a rule.

### What is not on `main`

`docs/ZERO_TRACKER_RECON.md` and `benchmarks/zero_tracker_recon_*_2026-08-09.*` (5 files). Every
number in them is from an **unsound probe** by construction, and the sound arm has since been
measured and landed. Reachable at `pull/481/head` = **`8c0428f98f27ac6ab8ee5cd663afdaf35753f08f`**;
indexed in `benchmarks/pr_triage_2026-08-13.meta`.

PR #482, opened 2026-08-10
The `&mut [u8]` recon-kernel refactor, WRITTEN — sound, lands on #481's unsound ceiling at all 8 cells, and +1.6 MB instead of +96 MB

lilith:
**Do not merge.** Companion doc: [`docs/MUT_RECON_KERNELS.md`](https://github.com/imazen/rav1d-safe/blob/perf/mut-recon-kernels/docs/MUT_RECON_KERNELS.md). Base `main` @ `cebb97f`.

PR #481 priced the `&mut [u8]` recon-kernel refactor and did not write it, and named a blocker: the geometry #474 chose costs **+96.3 MB at 8bpc / +191.0 MB at 10bpc**, which is not shippable. This writes the refactor on a **column-compact, one-superblock-row, per-worker** band instead, so both halves land in one pass.

## What is NOT done, first

* **Inter frames are not converted.** Scope is frames whose reconstruction is entirely intra (`KEY`/`INTRA_ONLY`). `mc`, OBMC, inter-intra blending and warp still write the shared picture through the tracker, and a frame containing them declines wholesale. Every writer of a recon plane has to move together — intra prediction, palette, CfL and the inverse transform all write the same pixels — so a subset conversion leaves the tracker on the path and measures null (#481's own argument for not writing a partial one). **228 `PicOffset` parameters remain** against the 39 converted: `src/mc.rs` 59, `looprestoration` 25+15+10, `cdef` 18+16+9+7, `safe_simd/mc*` 17+17+3, `loopfilter` 9+2+1, `lf_apply` 7, `filmgrain` 5, `recon.rs` 4 (the inter path), `lr_apply` 3.
* **The filter chain is not converted** — 47% of the post-#474 population, and no picture-buffer design reaches it. `src/ctx.rs:99` (2.53 M/frame) is bigger than everything here and is not a picture buffer at all.
* Also declining to today's path: `allow_intrabc`, `n_fc > 1`, `c-ffi`/`asm`, `__simd_test`, negative strides.
* Not measured: x86_64 (compile-checked only), `--features unchecked`, t=16, anything below 4K, any vector with loop restoration live.

## Soundness

The band is a plain `Vec` field of **`Rav1dTaskContext`** — the worker's own struct, already `&mut` in every recon signature — so `ReconPlanes::dst()` hands out a real `&mut [u8]` and exclusion is a **borrowck fact**, not a run-time record. No `unsafe` was reached for at any site. #474 could not do this: its buffers hang off `Rav1dFrameData`, which tile tasks reach through `fc.data.try_read()` as a SHARED reference, whose only expressible mutation is interior — i.e. the tracker.

## Registrations — the sound arm reaches the unsound ceiling exactly

`--features probe-sites`, `v4k_8tile` 8bpc, `lost=0`, same binary both arms (`RAV1D_OWNED_RECON=0` disarms):

| arm | t=1 | t=8 | distinct |
|---|---|---|---|
| band off (= `main`) | 7,924,706 | 22,700,725 | 54 |
| **band on** | **6,005,602** | **11,401,399** | 49 / 48 |

The off column reproduces #455's and #481's committed `base` census bit for bit. **The on column is #481's `ceil` arm, exactly** — which #481 obtained by composing #474 with `tracker: None`. Reached here from `main`, soundly, without #474, because the same change also removes the per-row split that took base from 7.9 M to 22.7 M at t=8. **11,299,326 registrations/frame gone at t=8 (49.8%).**

## Memory — the blocker, closed

Peak RSS, `/usr/bin/time -l`, 20 frames:

| arm | 8bpc t=8 | 10bpc t=8 |
|---|---|---|
| `main` | 106.2 MB | 157.2 MB |
| **band on** | **107.8** | **160.4** |
| #474 / #481 | 202.5 | 348.3 |

**+1.6 MB and +3.2 MB**, against +96.3 and +191.0 — 59× and 60× smaller. t=1 is +0.25 / +0.44 MB. The arithmetic: 8 workers × 960 columns × 128 rows × 3 planes ≈ 2.9 MB at 8bpc, allocated once and reused across superblock rows and frames.

## Wall clock

n=9 rounds, 288 rows, **`foreign_max = 0` on every one**, idle M4 Pro, `scripts/perf/verify_gap.sh`, two-point wall fit at 2 and 20 frames, rotating arm order, dav1d 1.5.4 `--framedelay 1` in the same interleaved sweep, no `nice` on timed runs. `head` and `headoff` are the SAME BINARY (`RAV1D_OWNED_RECON=0`).

Ratio to dav1d `--framedelay 1`:

| | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| `base` 8bpc | 1.296 | 1.574 | 1.627 | 1.873 |
| **`head`** 8bpc | **1.273** | **1.332** | **1.321** | **1.474** |
| *#481's UNSOUND ceiling* | *1.276* | *1.333* | *1.313* | *1.475* |
| `base` 10bpc | 1.442 | 1.621 | 1.658 | 1.915 |
| **`head`** 10bpc | **1.395** | **1.456** | **1.471** | **1.591** |
| *#481's ceiling, 10bpc* | *1.393* | *1.452* | *1.467* | *1.610* |
| whole-tracker ceiling (#467) | *1.160* | *1.264* | *1.262* | *1.345* |

**The sound arm lands on the priced ceiling at all eight cells** — within 0.008 at 8bpc, 0.019 at 10bpc, and marginally better at 10bpc t=8. 8bpc t=1 at **1.273** is under the ~1.30× bar.

Paired within-round `head/base`, exact two-sided sign test: **0.9823 / 0.8424 / 0.8134 / 0.7845** at 8bpc (p = 0.039 / 0.004 / 0.004 / 0.004), **0.9710 / 0.9004 / 0.8822 / 0.8279** at 10bpc (p = 0.004 each). Bands **disjoint at 7 of 8 cells**; 8bpc t=1 OVERLAPS [0.9699..1.0218] and is reported as unseparated rather than folded into the headline. Scaling t=1→t=8: 4.71× → **5.88×** (dav1d 6.80×).

### The honest negative: the seam costs the SHARED path 1.2–3.0%

`headoff` — same binary, band disarmed, i.e. every frame this change declines, which today is **every inter frame** — is consistently slower than `base`:

| `headoff` / `base` | t=1 | t=2 | t=4 | t=8 |
|---|---|---|---|---|
| 8bpc | 1.0115 | 1.0180 | 1.0193 | 1.0216 |
| 10bpc | 1.0198 | 1.0266 | 1.0301 | 1.0245 |

9/9 rounds slower at 7 of 8 cells (p = 0.004). The `ReconDst` branch on the tracked path is not free. Until the inter path is converted, a mixed-GOP stream pays 1–3% on every inter frame and collects the win only on key frames — which is the argument for treating this as the first half of a conversion, not a shippable end state.

### An earlier n=9 sweep was thrown away

It reported head at 9.13× dav1d at 8bpc t=8 with no thread scaling, and the band-DISARMED arm worse than the armed one — which no mechanism explains. The staged `bench_head` had been overwritten by a `--features probe-sites` build (cargo writes both feature sets to the same `target/release/examples/bench_ab_decode`). `sample` put `site_probe::record` at **51.4% of leaf samples**; `/usr/bin/time -l` had already shown 45.5 s user against base's 10.0 s. Rule out of it: `strings <binary> | grep site_probe` before staging an arm, and profile before believing a regression bigger than the mechanism.

## Gates

* Corpus set-diffed **BY NAME with the md5 as the value** against `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`: `--threads 1` **766 PASS / 768 keys / 0 differing**; `--threads 8` **753 PASS / 755 keys / 0 differing** after dropping the two #479 film-grain groups from both sides (`8-bit/film_grain`, `10-bit/film_grain`; `md5_inventory --skip-group` added, prints `skipped_groups=` on the TOTAL line).
* Frame md5 across `main`/band-on/band-off at t=1/2/4/8 × 8bpc/10bpc: **24 of 24 identical**.
* `mt_stress` 1/2/4/8/16 × 5 pass; `multi_decoder_pressure` 12 × 3 × {1,2,4,8,16} **PASS**.
* Liveness asserted before any timing: an `assert!(false)` planted after `ReconBand::arm` fires on the first superblock row.
* **The exclusion property proved the same way as `forbid`**: two live regions over one band planted in `owned_recon`'s tests → `error[E0499]: cannot borrow 'b' as mutable more than once at a time`. Restored byte-exact. There is deliberately no run-time test for it — one that passes on a string match would prove nothing.
* **`#![forbid(unsafe_code)]` proved NON-VACUOUSLY.** `src/owned_recon.rs` carries no `cfg` on its `mod` declaration, so it is compiled on the default feature set; a plant there is rejected at `lib.rs:13`, as is a second plant in `include/dav1d/picture.rs`. Both restored byte-exact by sha256 **and** `git diff --exit-code`. There is no compile-time feature to re-prove under — arming is a runtime switch so both A/B arms are one binary.
* Both standing hazards replanted under `--features __probe_wide` (the `4af62ae` in-lock `state` re-read deleted from `add_contended` **alone**, and `active()` cut to one shard) — `wide_exclusion` FAILS on each, green after each byte-exact restore. `crates/rav1d-disjoint-mut` is untouched by this branch.
* Miri: `crates/rav1d-disjoint-mut` is byte-identical to `cebb97f` on this branch (`git diff cebb97f -- crates/` is empty), so neither model can regress from a change that is not there. Both models are being re-run for the record anyway; the result will be posted as a comment.


lilith:
## Miri, and a real breakage the box could not have found

**Miri — both models green.** `crates/rav1d-disjoint-mut` is byte-identical to `cebb97f` on this branch (`git diff cebb97f -- crates/` is empty), so neither model can regress from a change that is not there; both were run anyway, each target in isolation via `--no-fail-fast`:

| model | result |
|---|---|
| Stacked Borrows | **61 passed, 0 failed**, no UB across 9 targets (`soundness` 992.7 s) |
| Tree Borrows (`-Zmiri-tree-borrows`) | **61 passed, 0 failed**, no UB across 9 targets (`soundness` 993.1 s) |

**The `asm` arms did not compile, and nothing runnable on this box could have said so.** `--features asm` cannot be built on macOS at all — `main` itself fails c-ffi's `EAGAIN` const-assert there (macOS 35 vs Linux 11), and the aarch64 asm leg wants a cross-assembler that is not installed. Checking against `--target x86_64-unknown-linux-gnu` found three real breakages, all mine:

* every `*_c_erased` FFI shim recovers a `PicOffset` from `FFISafe` and hands it to a Rust fallback that now takes `&mut ReconDst`;
* `pal_pred::Fn::call`'s asm arm still called `as_mut_ptr` on the enum;
* `ipred.rs` / `itx.rs` needed `Strided` back (cfg-gated) for the `.stride()` those arms take.

Fixed in `9492cfc`. Every edit is inside `#[cfg(feature = "asm")]`, so the default build is unchanged — re-gated anyway (corpus t=1 766 PASS / 768 keys / 0 differing, t=8 753 PASS / 755 keys / 0 differing).

Compile matrix now green: default on aarch64-darwin / x86_64-darwin / wasm32-unknown-unknown / aarch64-linux, `--features c-ffi` on aarch64-linux, `--features asm` on x86_64-linux.

**Full workspace suite** (`cargo test --release --workspace`): 30 targets, **0 failures**, including `decode_md5_verify` (14 tests) and `decode_permutations` (19 tests, 728 s).

PR #485, opened 2026-08-10
perf(lf-band): LfBlock::fill attributed 69/31, and the band that removes 29.6% of all registrations is REFUSED (default off)

lilith:
Campaign #455. Takes on `LfBlock::fill` — the decoder's largest guard site, 3,835,042
registrations/frame at `v4k_8tile` 8bpc t=8, 33.6% of 11,401,399 — and lands **one
instrument, one measured negative, and no perf change**. The default build is
behaviourally `main`.

## Read this first: what is NOT here

* **No speedup ships.** The band that produced the 3.7% below is **default OFF** because
  it fails 1–64 corpus vectors per run at `--threads 8`, nondeterministically.
* **The V pass's cheap sound win is measured but NOT taken** (§16: cap 4→32 divides its
  registrations by 1.971). It is blocked on a rectangular scratch and on decoupling the
  NEON kernel batch from the guard batch; both untouched.
* **`cdef_arm` (1,863,648) and `ctx.rs:99` (2,534,988) untouched**, as in every prior round.
* Not measured: x86_64/wasm32 (compile-checked), `asm`/`c-ffi`, `unchecked`, t=16, any
  vector below 4K, any vector with loop restoration live.
* **The full wall-clock sweep never ran.** The box was held by another agent's `measlock`
  for the window it needed, and once the band was refuted there was nothing shippable left
  to sweep. The only timed number here is the paired user-CPU A/B.

## 1. Attribution (new instrument, `--features __probe_lf_hist`)

The census gives this site one number; it cannot say which pass produces it. Split, at t=8
(reconciles to 3,835,042 **to the registration** after the probe's warmup scale, which is
the instrument's control):

| pass | regs/frame | share | shape |
|---|---|---|---|
| H — `filter_plane_cols_*` | **2,656,552** | 69.3% | `4*groups` rows × `2*reach` px |
| V — `filter_plane_rows_*` | **1,178,490** | 30.7% | `2*reach` rows × `4*groups` px |

79% of opens hit the `LF_BATCH_MAX = 4` cap exactly. Pricing the obvious lever from the
same run — what each natural run *would* have cost uncapped:

| cap 4 → 32 | now | uncapped | ratio |
|---|---|---|---|
| V | 1,178,490 | 597,876 | **1.971** |
| H | 2,656,552 | 2,656,552 | **1.000** |

Exactly 1.000 for H, structurally: its rectangle grows in the ROW direction, so a run of
`n` groups costs `4n` guards however it splits. **69.3% of the site cannot be bought by
fusing along the run at all.**

## 2. The band, and its registration table

A per-superblock **read cache** (`src/lf_band.rs`): one contiguous immutable guard per
picture row, once per superblock, for all of that superblock's calls. It needs none of the
halo handshake §11b found inexpressible, because it never owns anything — `close_band`
writes the picture with the same mutable guard `close` always took, and mirrors into the
band, so the band never diverges and any call may fall back to the picture path.

`--features probe-sites`, `lost=0`, one binary:

| | band OFF | band ON |
|---|---|---|
| whole decoder, t=8 | 11,401,399 | **8,026,355** (−29.6%) |
| `LfBlock::fill` | 3,835,042 | 35,570 (−99.1%) |
| `LfBand::fill_from` | 0 | 424,428 @ **134 B** |
| writes | 17,852 | **17,852** (identical) |

## 3. Why it is default OFF — measured, and the first run lied

**The read set is 2-D SPARSE; every per-row band is contiguous.** The difference is
columns and rows reserved but never read, and another worker is legitimately writing them:

```
current:  &     _[163840..163968]   <- the band's 128-px row copy-in
existing: &mut  _[163944..163952]   <- a concurrent 8-px write inside it
```

Same defect class as §11c's hull, transposed: the hull reserved the gaps between rows,
this reserves the gaps between edges. `md5_inventory --threads 8`, group `8-bit/data`
(358 vectors), repeated on one box:

| arm | pass | error |
|---|---|---|
| base `0f6bf10`, its own binary | 358 ×3 | **0 ×3** |
| column band only | 357 / 356 / 358 | **1 / 2 / 0** |
| column + row band | 294 | **64** |
| default build (band off) | 358 ×4 | **0 ×4** |

The column-only band's **first** full t=8 corpus run passed 753/755 `SETDIFF: CLEAN`, and
it was committed on that. One sample. A rare false positive is a decode *failure*, not a
wrong pixel, so an md5 diff cannot see it — only an error count can.

`tile_threading_active()` does not rescue it: that latch gates concurrent **tile** workers,
while this concurrency is between **sbrow filter tasks** (`src/thread_task.rs:1030-1043`),
which exist whenever `n_tc > 1`. Gated that way the t=8 census is byte-identical to `main`
(the band provably never armed) and `8-bit/data` **still produced 8 errors in one run of two.**

## 4. The number the armed arm is still good for

Paired user CPU, `v4k_8tile` 8bpc t=8, 20 frames, one binary, rotating arm order, under
`measlock`, n=9, foreign_max=1 (load-tagged):

**band / band-off = 0.9628 median, [0.9523..0.9781], 9/9 faster.**

Not available — but it is the first *direct* price for removing ~2.4 M of this site's
registrations without paying extent, where §11f could only bound it from above by doubling.

## 5. Gates (default build)

* Census `lost=0`: **6,005,602** t=1, **11,401,399** t=8 — both exactly §11a's numbers.
* Corpus set-diffed BY NAME with the md5 as the value vs
  `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`: t=1 **766 PASS / 768 keys /
  0 differing, CLEAN**; t=8 with `8-bit/film_grain` + `10-bit/film_grain` dropped from
  BOTH sides (#479) **753 PASS / 755 keys / 0 differing, CLEAN**. `8-bit/data` at t=8
  clean 4 of 4.
* Frame md5 identical across band {on,off} × t {1,2,8} × {8,10}bpc: 12/12,
  `a00c11f454328023c58af14d55544cff` / `4f218411bc6ee4cc9c630fe827337fa2`.
* **Teeth proved:** an off-by-one planted in the band read fails **13/14** corpus tests;
  restored byte-exact (sha256 + `git diff --exit-code`), green again after `touch`.
  **Honest gap:** deleting the `close_band` mirror is NOT caught (14/14 still pass) — the
  mirror is kept as an invariant, not because a test can see it.
* `#![forbid(unsafe_code)]` proved ACTIVELY in `src/lf_band.rs` (unconditionally compiled):
  `error: usage of an 'unsafe' block`, lint anchored at **`lib.rs:13`**, not a module-local
  `cfg_attr`. Restored byte-exact (sha256 `6512f6ea…ede22`, `git diff --exit-code` clean).
* `crates/rav1d-disjoint-mut` byte-identical to base (`git diff 0f6bf10 -- crates/` empty;
  `tracker_shard.rs` = `d4e03d4a…423176d`, the value #482/#483 recorded).
* Clippy: both CI legs clean on aarch64 **and** x86_64. `--all-targets` x86 fails on base
  and head alike, in files this branch does not touch.

## 6. For the next attempt

* **Do not build another contiguous band.** Column-major, row-major and both are now
  measured. The obstruction is that the read set is sparse in 2-D and the tracker's unit is
  an interval.
* **Take the V cap lift** (1.971×, no extent change): blocked only on `LF_BW`'s
  `assert!(LF_BW * LF_BW == LF_BLOCK_LEN)` and `loopfilter_arm.rs:156`'s
  `LF_GROUPS == LF_BATCH_MAX`. Watch `LfScratch::new`'s zero-init — ~100 K calls/frame.
* **H needs a different mechanism, not a bigger batch** (cap-lift ratio exactly 1.000).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,6 +87,11 @@ probe-sites = ["rav1d-disjoint-mut/__probe_sites"]
 # Absent from `default` and from every published feature. See
 # `src/loopfilter.rs::lf_hull_reads` and docs/MUT_RECON_KERNELS.md section 10.
 __probe_lf_hull = []
+# Measurement-only: counts `LfBlock::fill`'s registrations split by direction
+# (H = vertical edges, `filter_plane_cols_*`; V = horizontal edges,
+# `filter_plane_rows_*`) and by run length, so the site's population can be
+# attributed before anything is built against it. Counts only, never a policy.
+__probe_lf_hist = []
 probe-tinynop = ["rav1d-disjoint-mut/__probe_tinynop"]
 probe-addnop = ["rav1d-disjoint-mut/__probe_addnop"]
 probe-noscan = ["rav1d-disjoint-mut/__probe_noscan"]
--- a/docs/MUT_RECON_KERNELS.md
+++ b/docs/MUT_RECON_KERNELS.md
@@ -1090,3 +1090,114 @@ Net **-2,459,800 registrations/frame, -21.6% of the whole decoder's
 population**, bought at an extent of 142 contiguous bytes per guard. That is
 the opposite trade from §11c's hull, which removed 3.46 M by paying 50-60 KB
 of strided extent and measured **2.65x SLOWER**.
+
+## 18. The band is REFUTED under concurrent filtering — and the first run passed
+
+§17's registration table is real and the band is nevertheless **not
+shippable**. It ships **default OFF** (`RAV1D_LF_BAND=1` to arm), in the shape
+`__probe_lf_hull` already uses for the strided-hull negative.
+
+### What went wrong
+
+**The loop filter's read set is 2-D SPARSE, and every per-row band is
+contiguous.** A band reserves, for each of its rows, one span covering the
+whole superblock; the pass actually reads only the `+-reach` tap windows
+around the edges that filter, and a row in which nothing filters is not read
+at all. The difference is columns and rows that are reserved but never read —
+and under concurrent filtering another worker is legitimately writing them:
+
+```text
+current:  &     _[163840..163968]   <- the band's 128-px row copy-in
+existing: &mut  _[163944..163952]   <- a concurrent 8-px write inside it
+```
+
+That is the SAME defect class as §11c's hull, transposed. The hull reserved
+the gaps BETWEEN rows; this reserves the gaps BETWEEN edges. §11c's version
+was merely slow because its extent hit the wide path; this one is a false
+positive, i.e. a decode failure.
+
+### The numbers, and the sample that lied
+
+`examples/md5_inventory --threads 8`, group `8-bit/data` (358 vectors), same
+box, same load, repeated:
+
+| arm | pass | error |
+|---|---|---|
+| base commit `0f6bf10`, its own binary | 358 / 358 / 358 | **0 / 0 / 0** |
+| band off (this binary) | 358 / 358 | 0 / 0 |
+| **column band only** | 357 / 356 / 358 | **1 / 2 / 0** |
+| **column + row band** | 294 | **64** |
+| default build (band off), after the fix | 358 x4 | **0 x4** |
+
+**The column-only band's first full `--threads 8` corpus run passed 753/755
+`SETDIFF: CLEAN`.** One sample. It was committed on that basis, and the
+failure only appeared when the row band widened the same defect enough to fire
+reliably. "766 vectors passed" is evidence, not a proof — and a rare
+false positive is a decode failure, not a wrong pixel, so it cannot be
+detected by an md5 diff at all, only by an error count.
+
+### `tile_threading_active()` is the WRONG gate, measured
+
+The obvious rescue is the latch that already lets `fill_hull`, `block_mut` and
+`compact_read` widen a reservation. It does not work, because it gates
+concurrent **tile** workers and the loop filter's concurrency is between
+**sbrow filter tasks** — inserted for `sby+1` before the selected task for
+`sby` even runs (`src/thread_task.rs:1030-1043`) — which exist whenever
+`n_tc > 1` however many tiles a frame has. Gated that way:
+
+* the `v4k_8tile` t=8 census is byte-identical to `main` (11,401,399), so the
+  band provably never armed on that vector, and
+* `8-bit/data` at t=8 **still produced 8 errors in one run of two.**
+
+A gate that is right for this site would have to mean "no other thread can be
+filtering this picture", which the decoder does not currently expose.
+
+### What the default build is
+
+Byte-for-byte `main`'s behaviour, verified rather than assumed:
+
+* census, `--features probe-sites`, `lost=0`: **6,005,602** at t=1 and
+  **11,401,399** at t=8 — both exactly §11a's numbers;
+* corpus set-diffed BY NAME with the md5 as the value, against
+  `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst`: `--threads 1`
+  **766 PASS / 768 keys / 0 differing, SETDIFF: CLEAN**; `--threads 8` with
+  both film-grain groups dropped from BOTH sides (#479) **753 PASS / 755 keys
+  / 0 differing, SETDIFF: CLEAN**;
+* the write population is untouched in every arm: 17,852 per frame, split
+  17,327 + 525 between `close` and `close_band` when armed.
+
+### The one number the armed arm is still good for
+
+Paired user CPU, `v4k_8tile` 8bpc t=8, 20 frames, one binary, arms interleaved
+with rotating order within each round, under `measlock`, n=9 (foreign_max=1,
+so load-tagged):
+
+| | median | band | faster |
+|---|---|---|---|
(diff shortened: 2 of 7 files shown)


lilith:
### Gate results that landed after the PR body was written

* **Miri, Stacked Borrows: 61 passed, 0 failed, `rc=0`** — the same 61/61 §14 recorded, re-run on this tree rather than inherited. Tree Borrows was still running at hand-off; an earlier run of the same command on this branch (commit `07d992f`) returned `rc=0` with 0 failures. `crates/rav1d-disjoint-mut` is byte-identical to base on every commit here (`git diff 0f6bf10 -- crates/` empty, `tracker_shard.rs` = `d4e03d4a…423176d`), so neither model can regress from a change that is not there — but the SB number above is a real re-run, not that argument.
* **`cargo test --release`**: 75 + 6 + 6 + 3 + 14 passed, **0 failed** across the targets that had finished (incl. `decode_permutations`, 275 s, and `decode_md5_verify`). The run had not reached its last targets at hand-off.

### Gates NOT run — do not read the above as a full pass

* `cargo test --lib` in the **DEBUG** profile (queued behind the release run; #482 hit a profile-dependent panic on its Coverage leg, so this one matters).
* `scripts/perf/multi_decoder_pressure.sh` and a standalone `mt_stress` 1/2/4/8/16 sweep. `mt_stress` is inside `cargo test --release`; the 12-process pressure script was not run separately.
* The **standing-hazard replant** under `--features __probe_wide` (delete the in-lock `state` re-read from `add_contended`; cut `active()` to one shard; confirm `wide_exclusion` FAILS each time, restore byte-exact). Not run. `crates/` is byte-identical to base, so the hazards are structurally the ones #483 verified, but that is an argument and not this branch's measurement.
* The **full wall-clock sweep** (n≥9, 8 cells, dav1d interleaved, min/max bands, DISJOINT verdicts, RSS). Never started: the box's `measlock` was held by another agent for the window it needed, and once the band was refuted there was no shippable arm left to sweep. The default build's registration census is byte-identical to `main` at t=1 and t=8, so there is nothing here a wall sweep could move.

Given the default build is behaviourally `main` (census identical, both corpora `SETDIFF: CLEAN`), the value of this PR is the instrument and the refutation, not a perf delta.

lilith:
## Closing: the band is refuted on `main` as a MECHANISM, and the instrument now predicts its failure

Triage of the 15 open PRs, 2026-08-13. This round shipped no perf change by its own decision — the
band is **default OFF** because it fails 1-64 corpus vectors per run at `--threads 8`,
nondeterministically — so there is nothing here to merge.

Its refutation is on `main` in a stronger form than a note: it is **built into the instrument**.
`crates/rav1d-disjoint-mut/src/bounds_probe.rs` names this round in its module header as one of the
"count reduction" attempts it exists to have predicted, and computes the band as a standing
counterfactual:

* `:328` — "which is exactly the shape #485's loop-filter band took";
* `:365` and `:891` — "Counterfactual (#485's band): widening this guard to the full picture rows…";
* `include/dav1d/picture.rs:310` and `:367` — the three widening attempts (#469 rectangle, #475 hull,
  #485 band) named together at the site that must not repeat them;
* `docs/BOUNDS_MAP.md` — **"#485 BLOCKED at 2..16 per 1406 frames against its measured 1, 2, 0"**.
  That is the map's own correctness gate passing: it retrodicted this round's error counts, from the
  widening budget, without being told about them.

The mechanism verdict this round wrote is the one `main` acts on: **the read set is 2-D SPARSE and
every per-row band is contiguous**, so the reserved-but-never-read columns are where another worker
is legitimately writing. And its instruction — *"Do not build another contiguous band. Column-major,
row-major and both are now measured"* — is honoured: what shipped instead is an **exact strided
rectangle** (#505, default since #513), which reserves no slack at all.

Its two other conclusions also stand and are on `main`:

* the H/V split of `LfBlock::fill` (69.3% / 30.7%) and **H's cap-lift ratio of exactly 1.000** —
  `AGENT_BRIEF.md` §6, "structurally null … H's rectangle grows in the ROW direction";
* the V cap's 1.971x, which #488 then built and measured as no wall-clock win.

### One thing to carry forward that is easy to lose

> "The column-only band's **first** full t=8 corpus run passed 753/755 `SETDIFF: CLEAN`, and it was
> committed on that. One sample. A rare false positive is a decode *failure*, not a wrong pixel, so
> an md5 diff cannot see it — only an error count can."

That is the sharpest statement of a trap this campaign keeps hitting, and it was self-reported.

### What is not on `main`

`docs/MUT_RECON_KERNELS.md` §16-§18 as this branch wrote them (that file has moved substantially
since), `src/lf_band.rs`, and the **0.9628 median paired user-CPU** for the armed arm — the first
*direct* price for removing ~2.4 M of this site's registrations without paying extent, 9/9 faster,
load-tagged. That number is worth knowing about even though the arm is unsound, because #505 has
since shown the cost model it bears on: cost tracks **distinct shard lines visited**, not records
filed, and `LfBlock::fill` files 8.98 records per 2.09 lines — 31.7% of the population for 3.9-4.4%
of tracker CPU. So this site being "33.6% of registrations" is **not** a priority ordering, which is
the framing correction a future reader of this branch most needs.

Reachable at `pull/485/head` = **`d158217503592f0faffb3bb21f73666daa97adc4`**; indexed in

PR #488, opened 2026-08-10
lf V-batch cap: 1.971x fewer registrations, measured as no wall-clock win — plus ctx.rs:99 attributed

lilith:
Two bounded tasks from #455. **Neither shipped a perf change, and both produced
a number.** What is in the diff is instruments, records, and two measured
negatives; `src/loopfilter.rs` and `src/safe_simd/loopfilter_arm.rs` are back to
`f87b12c` and the diff against base is **pure addition** (no removed lines, the
kernel file byte-identical).

## Win 1 — the V pass's batch cap: built, correct, and NOT worth it

`docs/MUT_RECON_KERNELS.md` §19.

The count came out exactly as PR #485 priced it — **1.971x**, V 1,178,490 ->
597,876 per frame, whole decoder 11,401,399 -> 10,820,785, and the delta is
*only* the V pass. The wall clock did not follow:

| 8bpc | ratio (head/base) | rounds faster | p | disjoint |
|---|---|---|---|---|
| t=1 | 1.0157 | 1/8 | 0.070 | no |
| t=2 | 1.0162 | **0/8** | **0.008** | no |
| t=4 | 1.0150 | 3/8 | 0.727 | no |
| t=8 | **1.0005** | 4/8 | 1.000 | no |

10bpc lands in [0.993, 1.003] at every thread count and is never significant.
n=7-8, paired per-round ratios, arms interleaved with rotating order, **loaded
box** (`measlock --load-ok`, `foreign` 14-24) — so ratios only, no absolutes.

**The number that should have been computed first, in two lines, before any
code:** 580,614 registrations x #485's own measured **4.04 ns** each = 2.35
ms/frame = **<= 0.51% of a t=8 frame's CPU**. `1.971x` is a ratio on 30.7% of
ONE site; nobody had converted it into milliseconds, and it is at or below what
this box resolves.

**It was sound**, and that argument is kept in the docs because it is the useful
part. A fused run's rectangle is exactly the union of its members' — every one
of its `4 * groups` columns belongs to a group that filters, at the same `wd`
and therefore the same `2 * reach` rows — so unlike #485's band there is no
slack for a concurrent write to fall into. The one-line test that separates all
three schemes: *does the reservation contain a byte no member of the batch
reads?* Hull: yes. Band: yes. Fused run: no.

**Where the time actually went.** The first cut measured **+3.0% t=1 / +7.9%
t=8**, and an isolation arm (whole machinery, cap pinned back to 4) was
**+18.7% at t=1** — so the cost was the MACHINERY, not the wider runs. A
`sample` profile named the leaves; three fixes (per-direction threshold table,
run-own const-generic scratch stride, no chunk loop when `w <= 16`) took it to
1.014 / 0.981. The residual ~2% is the thread-local the 2,048-pixel scratch
forced, which is four times the whole prize. **§11 said price the extent, not
just the count. This adds: price the machinery, and price the count in
milliseconds before building.**

The implementation is preserved at 61f88dc + 362e5d9 (parity 4/4 across
8/10/12bpc at every group count to 32, corpus 766/766 and 753/753, frame md5
identical) so it can be resurrected if the tracker ever gets more expensive.

## Win 2 — `ctx.rs:99`: attributed, mechanism named, half of it priced

`docs/MUT_RECON_KERNELS.md` §20, `docs/OWNERSHIP_MODELS.md` §7e,
`benchmarks/ctx99_sites_2026-08-10.tsv`. **Not reduced** — say so first.

A cfg-gated `#[track_caller]` on `CaseSetter::set_disjoint` (probe builds only)
splits the site across its 12 real callers, which **sum to 2,534,988 exactly**.
All 12 are mutable, mean extent 2.3 bytes. The count is structural:
`BlockContext` is ~20 separate `DisjointMut<Align8<[T; 32]>>` fields and one
`CaseSet::many` updates 8-13 of them for the same block at the same offset.

Every call is over exactly two directions, so it splits **50/50**:

* `f.a[t.a]` — behind the shared `try_read()` guard: §1's blocker verbatim, and
  its cross-tile disjointness is the tile-keyed argument §3 rejected. Not
  reducible by ownership.
* `t.l` — a per-worker field of `Rav1dTaskContext`, already reached through
  `&mut` wherever `recon_band` is. **1,267,494/frame, 11.1% of the decoder's
  population, removable with borrowck as the proof and no `unsafe`;**
  `DisjointMut::get_mut(&mut self)` already exists.

What blocks it is shape: `CaseSet::many` takes a homogeneous `[T; N]`, so both
directions must be the same type. Converting means splitting all 22 `CaseSet`
sites that mention `t.l`, or a macro that expands each body twice. Not attempted
here. **Do not attack it by coalescing fields** — that fails §19d's one-line
test.

## Gates (all on the final tree, `e19c4f4`/`7c2d1f3`)

* Corpus **by name, md5 as the value**: `--threads 1` 766 PASS / 768 keys / 0
  differing **SETDIFF: CLEAN**; `--threads 8` with `8-bit/film_grain` +
  `10-bit/film_grain` dropped from BOTH sides (#479) 753 / 755 / 0 **CLEAN**.
  Re-run after the last code-touching commit, not before.
* Census reproduces base exactly: **11,401,399** per frame, `lost=0`.
* Frame md5 identical to base at 8bpc and 10bpc, t=1 and t=8.
* `#![forbid(unsafe_code)]` proved **actively** in an always-compiled file with
  no module-local attribute (`src/ablate.rs`): the error anchors on **lib.rs:13**
  and overrules `#[allow(unsafe_code)]`. Restored byte-exact (sha256 +
  `git diff --exit-code`).
* Teeth planted and confirmed failing, then restored byte-exact: the NEON parity
  test (wrong threshold index; wrong V stride at groups=9), the corpus
  (`copy_row_v` tail dropped: 13/14 tests FAIL), and the probe itself (break
  `note_natural` -> `LFCAP` cap-32 collapses 1.971 -> 1.000).
* Standing hazards under `--features __probe_wide`, run even though `crates/` is
  untouched (verified: empty diff, `tracker_shard.rs` sha
  `d4e03d4a…`): control passes; **deleting the in-lock `state` re-read** FAILS;
  **`active()` cut to one shard** FAILS. Restored byte-exact.
* Miri, **both models**, each target in isolation, `--no-fail-fast`: Stacked
  Borrows **61 passed / 0 failed, rc=0, no UB**; Tree Borrows **61 passed / 0
  failed, rc=0, no UB**.
* `cargo test --lib` in **debug** as well as release: 75 passed both.
* `mt_stress` 1/2/4/8/16 (25 cells) and
  `scripts/perf/multi_decoder_pressure.sh` (12 concurrent decoders, all md5s
  match the serial reference).
* **x86 clippy reproduced locally** (`--target x86_64-apple-darwin
  --all-targets --keep-going -D warnings`): 11 failing sites, set-diff against
  base **empty in both directions**. That form is stricter than CI, whose
  Clippy job omits `--all-targets` and so never lints `examples/`; all 11 sites
  are in `_dev` examples plus one test-only `fn`, and CI's Clippy job is green
  on both base and head.
* **CI dispatched on the branch** (run 31382794525): Clippy, Format, Coverage,
  SIMD permutation tests on both arches, and all four cross-compiles green;
  conformance / per-platform build-and-test still finishing at hand-off. Note
  the `pull_request` trigger did not fire for this PR by itself — it had to be
  `workflow_dispatch`ed.

## Known gaps

* The 9th sweep round was lost: two of my sweeps restarted from round 0 and ran
  concurrently after `measlock`'s owner-file format changed under a live holder.
  Committed data is the pre-restart snapshot at n=7-8.
* A cap-8 variant (stack scratch, no thread-local) was built and verified but
  deliberately **not** measured — its prize is <= 0.32% of t=8 CPU, further below
  the resolution than the arm that already measured null.
* Planting the un-chunked write-back (a real over-reservation) passed 358/358 at
  t=8 in one run, and that arm's liveness on the corpus vectors was not proved,
  so the null bounds nothing.
* Not measured: x86_64/wasm32 runtime, `asm`/`c-ffi`, `unchecked`, t=16, any
  vector below 4K.

Do not merge without a maintainer read of the revert rationale.


lilith:
**CI green: 21/21 jobs, run [31382794525](https://github.com/imazen/rav1d-safe/actions/runs/31382794525)** — including both conformance-permutation jobs, `windows-11-arm`, `macos-15-intel`, all four cross-compiles, and the threading race gates.

**Miri is now complete in both models**, each target in isolation, `--no-fail-fast`: Stacked Borrows **61 passed / 0 failed, rc=0**, Tree Borrows **61 passed / 0 failed, rc=0**, no UB reported in either. (The body previously recorded Tree Borrows as incomplete; it finished after that was written.)

One process note for whoever triages: **the `pull_request` trigger did not fire for this PR on its own** — CI had to be `workflow_dispatch`ed against the branch. Other open PRs on `perf/*`-style branches did get it automatically, so this looks like a one-off rather than a config gap, but it is worth a glance.

lilith:
Cross-ref from the bounds map's phase 2 (#489, `docs/BOUNDS_MAP.md` Part 2): the
map **contradicts §19d's soundness test**, though not this PR's conclusion — you
reverted the change, and the map agrees the site is not worth attacking.

> *does the reservation contain a byte no member of the batch reads?*
> Fused run — **no** (sound).

The premise is right; the conclusion does not follow, because `DisjointMut`
compares **one live reservation against another**, not a reservation against a
read set. Slack-freedom rules out #485's failure mode (a foreign write in a gap
nothing reads); it does not stop a foreign write landing in a byte the run *does*
read, and `fill_v`'s live reservation goes from 16 px to **128 px** while the
site's measured clearance to the nearest concurrent writer is **60 B** — and that
writer is `loopfilter.rs:887:14`, the loop filter's own write-back, not CDEF at
232 B (the 232 was a per-pair figure quoted as the site's; corrected in #489).

Budget column for `loopfilter.rs:710:14`, `--features __probe_bounds`
(#486): `k<=64` = 2, `k<=256` = 16 over 1406 frames of `8-bit/data`, so a +112 B
widening predicts 2..16 collisions, x the V pass's 30.7% share = **0.6..4.9 per
corpus pass**. Zero on either 4K vector.

**Your 753/753 at t=8 is evidence against that prediction and it is stated first
in #489.** What stops it settling the question is your own honest gap: the
planted un-chunked write-back — a genuine over-reservation — measured 358/358
clean, and #485 measured 1, 2 and **0** on three passes of that group. One pass
cannot resolve a ~1-5-per-pass rate. The direct measurement is one
`--features __probe_bounds` corpus pass on top of `61f88dc`: `mutable_overlaps`
must stay 0.

One thing neither round settled, and it matters because the two readings need
opposite fixes: a collision here is necessarily on a byte the run DOES read, so
it is either **a true ordering hazard** (the fused run reads all 32 tap windows up
front, possibly before their writers finish — wrong pixels, no panic, in an
`unchecked` build) or a spurious panic. Which one depends on the sbrow-filter
task's ordering contract.

Also landed in #489: a standing gate that fails the moment anything widens a
picture-plane reservation past its measured per-file ceiling — `loopfilter.rs` is
pinned at 32 B, so this change would now fail `cargo test --test mt_stress` in
debug, and CI's new `extent-gate` job in release. Mutation-proved both ways.

lilith:
## Closing: data landed in #518; the conclusions were already on `main` in four places

Triage of the 15 open PRs, 2026-08-13. This PR ends **"Do not merge without a maintainer read of the
revert rationale"**, and the rationale reads clean: `src/loopfilter.rs` and
`src/safe_simd/loopfilter_arm.rs` are back to `f87b12c`, the kernel file byte-identical, and the diff
against base is pure addition. Two measured negatives and two instruments.

### What landed (#518, merged `71a0fbd`)

The six data files, because `AGENT_BRIEF.md` §6 cites one of them **by name**
(`benchmarks/lf_vbatch_iso_2026-08-10_v1.tsv`) and `main` did not have it — the same dangling-citation
class #514 cleared for `TILED_SCALING.md` and `X64_APPLICABILITY.md` §F, and the one it did not
reach. `git merge-base --is-ancestor 97168c2 origin/main` passes.

* `lf_vbatch_2026-08-10_v1.tsv` (208 rows) + `_cap32.tsv` (248) — the paired A/B behind the t=8
  ratio of 1.0005.
* `lf_vbatch_iso_2026-08-10_v1.tsv` (18) + `_v3.tsv` (18) — **the isolation arm, which is the
  transferable result**: machinery kept, cap pinned back to 4, **+18.7% at t=1**. That is the row the
  brief cites, and it is the evidence for "price the MACHINERY a count reduction needs", not just the
  count.
* `lf_cap_census_2026-08-10.txt` (101) — the `LFCAP` census behind H's exactly **1.000** at caps
  4/8/16/32/64.
* `ctx99_sites_2026-08-10.tsv` (24) — the `#[track_caller]` split of `src/ctx.rs:99` across its 12
  real callers, summing to **2,534,988 exactly**. `main` cites that total in three places
  (`MUT_RECON_KERNELS.md` :27, :471, :521) and had never carried the breakdown, including the 50/50
  split whose `t.l` half is 1,267,494/frame of per-worker buffer.

`scripts/perf/paired_ab.py` came with them. The probe wiring in `src/ctx.rs`, `src/loopfilter.rs`,
`examples/probe_tracker.rs` and `lib.rs` is code and stays at `pull/488/head` =
**`419d2715ef0ed92dfe9432fde8741b954434a52b`**.

### The conclusions were already in four places, unchanged by this

* `AGENT_BRIEF.md` §6 — two rows: H's structurally-null 1.000, and the V cap's +3.0% t=1 / +7.9% t=8
  with the cost identified as the **machinery**, not the batch;
* `OWNERSHIP_MODELS.md:344` — the 1.971x count cut with t=8 ratio 1.0005, p=1.000;
* `SHARD_GRANULARITY.md:268` — cited as the converse of #500 (a cost change with **no** count change);
* `BOUNDS_MAP.md:473` — a whole section **CONTRADICTING** this round's soundness TEST (not its
  conclusion, which reverted the change anyway): *"does the reservation contain a byte no member of
  the batch reads?"* is not the collision criterion, because `DisjointMut` compares live
  reservations. Predicted 0.6..4.9 collisions per corpus pass against this round's one clean 753/753
  — **not confirmed, not refuted**, and this round's own planted over-reservation also measured
  358/358 clean, so one pass cannot resolve it. That is stated on `main` as an open settling
  measurement, and this round said the same in its own "known gaps".

### The line that should outlive the PR

> **The number that should have been computed first, in two lines, before any code:** 580,614
> registrations × #485's own measured 4.04 ns each = 2.35 ms/frame = **≤ 0.51% of a t=8 frame's
> CPU**. `1.971x` is a ratio on 30.7% of ONE site.

One framing correction for a future reader of this branch: #505 has since shown that cost tracks
**distinct shard lines a worker visits**, not records filed — `LfBlock::fill` files 8.98 records per
2.09 lines and is 31.7% of the population for 3.9-4.4% of tracker CPU. So "this site is 33.6% of
registrations" is not a priority ordering, and the ≤0.51% arithmetic above was, if anything,
generous to the lever.

Indexed in `benchmarks/pr_triage_2026-08-13.meta`.

PR #489, opened 2026-08-10
bounds map phase 2: the per-site VERDICT table, a standing extent gate with teeth, and the widening it contradicts (#455)

lilith:
Phase 2 of the bounds map (#455, stacked on #486). The map measured; this turns
it into three standing artefacts. **Nothing was coarsened or narrowed** — the
deliverable is the decision table, the gate that stops it going stale, and a
ranked list the next round picks from.

## What is NOT done, first

* **No candidate implemented.** Deliberate: a half-built candidate would muddy
  the artefact.
* **The V-batch contradiction is a PREDICTION with one clean corpus pass against
  it.** See below. Not confirmed, not refuted.
* **The standing gate's tight ceilings cover 3 of 11 picture-plane files.** The
  other 8 legitimately copy whole rows, so they are held only to the one-row
  rule; a 128-byte in-row widening introduced in `lf_apply.rs` or
  `cdef_apply.rs` would pass. `loopfilter.rs` has the tight entry because that
  is where both #485 and the V-batch live.
* **Not measured for the gate:** 12-bit, `multi-bit`/`oss-fuzz`, frames past the
  4th of each vector, x86_64, wasm32, `asm`/`c-ffi`, `unchecked`, t != 8.
* **The map's own 1.2% `UNMEASURED` hole is still open** — named and sized as
  candidate 6, not closed.
* **The whole-corpus `md5_inventory --threads 8` still cannot finish**, on this
  branch or on its parent. Pre-existing (#479); A/B'd, see the gates section.

## 1. `docs/BOUNDS_MAP.md` Part 2 — the verdict table

Every site, sorted by registrations/frame, with reservation, over-reservation and
its shape, whether its footprint ever conflicts concurrently, and one of
**NARROWABLE / COARSENABLE-UP-TO-N / BLOCKED**. Rules stated and applied
mechanically by `scripts/perf/bounds_verdicts.py` — deterministic, no new
decoding, re-runnable against #486's raw tarball. Record:
`benchmarks/bounds_verdicts_2026-08-10.{tsv,meta}`.

Shares over the corpus cell's 936,867,496 registrations at t=8:

| class | registrations | share |
|---|---|---|
| COARSENABLE-N (finite budget) | 407,958,230 | 43.5% |
| COARSENABLE-INF (no concurrent writer, EVER) | 288,180,835 | 30.8% |
| BLOCKED (butt-adjacent write) | 229,489,855 | 24.5% |
| UNMEASURED (instrument declined) | 11,238,576 | 1.2% |

### The map's own correctness gate — all three known refutations are marked

| | | |
|---|---|---|
| **#469 rectangle** | the ROWBAND counterfactual IS its record->hull widening | **BLOCKED** — 18,196 acquisitions would hit a concurrent foreign reservation over 1406 corpus frames, 7,612 a declared foreign FOOTPRINT |
| **#475 hull** | six declared-geometry sites at t=1 | **OVER-RESERVED 336x-1680x**, `lead` and `tail` 0.00 at every one — entirely inter-row gap. (Its LF row-band is 0, i.e. it would not collide; #475 was a COST refutation at 2.65x. Both axes have to be read, and the map says so.) |
| **#485 band** | +124 B lands between `k<=64` (2) and `k<=256` (16) | **BLOCKED** — 2..16 predicted per 1406 frames; it measured 1, 2 and 0 |

### Two corrections the full data forced

* **`loopfilter.rs:710:14`'s nearest concurrent writer is
  `loopfilter.rs:887:14` at 60 B** — the loop filter's own write-back in another
  sbrow filter task, 1,176,771 co-live pairs. `cdef_arm.rs:622:9` at 232 B is
  the *third* nearest; `BCONC`'s site-level `min_gap_mut` said 60 all along.
  Fixed in `OWNERSHIP_MODELS.md` §7d and `AGENT_BRIEF.md`; erratum added to
  #486's `.meta` rather than rewriting the record.
* **205 of 253 sites never execute on either 4K gap vector**, and only 49.5% of
  corpus registrations land at a site the 4K vector runs at all. `src/mc*` is
  **0** there — 33.0% of the corpus population — because `photo_4k.avif` is an
  all-intra still. The known "LR is off in both 4K vectors" is a subset of this.

## 2. Standing assertions — `include/dav1d/picture.rs::note_pic_extent`

At the one funnel every tracked picture-plane reservation passes through:

1. **may not span more than ONE picture row** (the inter-row gaps belong to other
   columns of the same rows, and AV1 tiles partition a frame BY COLUMN) — #469's
   and #475's defect;
2. **may not exceed its file's `PIC_EXTENT_CEILINGS` entry**, or one picture row
   where there is none — #485's, and the V-batch's.

**Why extent and not "reserved <= footprint":** at t=8 the ratio is already 1.000
at every hot site, and two of the three refuted attempts kept
`reserved == footprint` while widening. A ratio assertion would have passed #485
AND the V-batch.

Ceilings measured by the check's own counters over **575,577,925** reservations
across 1,734 frames (committed crash vectors + `8-bit/data` + `10-bit/data`,
t=8). `MAX_ROWS_TT = 1` over all 575 M — the shipped decoder never reserves
across a row boundary under tile threading. `src/loopfilter.rs` maxes at exactly
**32 B** over 90,170,282 registrations (`LF_BW = 16` px x 2 B).

**Cost: none.** Compiled under `debug_assertions` or `--features probe-sites`
only; the default release build has no counter, no atomic load and no branch
there.

**Coverage:** plain `cargo test` (debug) carries it inside the decoder;
`tests/guard_extent_budget.rs` runs it in release with three liveness assertions
(evaluated at all / evaluated **under tile threading** / whole-component
exemption exercised); new CI job **`extent-gate`** runs the corpus leg, which is
caller-gated by `RAV1D_EXTENT_GATE_CORPUS` — the test body never decides to skip.

### Teeth, by mutation

| plant | result |
|---|---|
| baseline | ok |
| `--features __probe_lf_hull` + `RAV1D_LF_HULL=1` (#475's shape, **no source edit** — the switch is already in the tree) | **FAILED**: 17 in-decoder panics, `loopfilter.rs:769:33 took a 1924 B picture-plane reservation ... spans 16 rows`, plus `MAX_ROWS_TT` at `left: 16, right: 1` |
| LF per-row read `W` -> `W + 96` (the V-batch's shape, **inside one row**) | **FAILED**: `loopfilter.rs:710:14 took a 100 B picture-plane reservation ...; the measured ceiling for that file is 32 B` |
| an `unsafe` block in `tile_threading_active()` | **FAILED to compile**, anchored on `lib.rs:13 forbid(unsafe_code)` |
| **DEBUG**, `cargo test --test mt_stress`, default features, no probe | ok (98.75 s) |
| the same with `W` -> `W + 96` planted | **FAILED** — `rav1d-worker-0 panicked at src/loopfilter.rs:710:14: took a 110 B picture-plane reservation ...; the measured ceiling for that file is 32 B` |
| after restore | ok, `git diff --exit-code` clean |

**The mutation found a hole in the gate**, which is why it is worth running: on
the first pass mutation 2 fired the in-decoder panic but NOT the test's own
`<-- OVER` flag, because that check was `bytes > ceiling && rows > 1` — so a
widening confined to one row, the V-batch's exact shape, was invisible to the
assertion meant to restate the invariant. Fixed in `7a5a910`; both halves now
apply the same bound. Only a mutation shows you which half is wrong.

All restores `sha256`-verified + `git diff --exit-code` clean, re-run after
`touch` per the mtime trap.

## 3. CONTRADICTED: the V-pass fused run's soundness TEST (#488 §19d)

**#488 already reverted the change** (no wall-clock win: t=8 1.0005, p=1.000). It
keeps the soundness argument deliberately, and that is what is contradicted:

> *does the reservation contain a byte no member of the batch reads?*
> Fused run — **no** (sound).

The premise is right — a fused run is slack-free. The conclusion does not follow:
**`DisjointMut` compares one LIVE reservation against another, not a reservation
against a read set.** `fill_v`'s live reservation goes 16 px -> 128 px against a
**60 B** measured clearance; the budget column says 2..16 collisions per 1406
corpus frames (x the V pass's 30.7% share = 0.6..4.9), and **0** on either 4K
vector.

**Evidence against, stated first:** #488 reports corpus 766/766 at t=1 and
753/753 at t=8 on the implementation. **Why that does not settle it:** the same
PR planted a *genuine* over-reservation and measured 358/358 clean on the same
group, and #485 measured 1, 2 and **0** on three passes of it. One pass cannot
resolve a ~1-5-per-pass rate. Settling measurement: one `--features
__probe_bounds` corpus pass on top of `61f88dc`.

**And the thing neither round established:** a collision here is necessarily on a
byte the run DOES read, so it is either a true ordering hazard (the fused run
reads all 32 tap windows up front, possibly before their writers finish — wrong
pixels, no panic, in an `unchecked` build) or a spurious panic. That depends on
the sbrow-filter task's ordering contract. The two readings call for opposite
fixes; settle it before the next attempt.

Joint conclusion of both rounds: **this site is not worth attacking on extent.**
60 bytes of clearance and a prize under the noise floor.

## 4. The ranked candidate list (none implemented)

1. **The 31.4% that can never collide** — 293,787,973 corpus registrations at
   sites with **no concurrent foreign write at any distance**, led by
   `mc.rs:121:61` (181,371,944 = 19.4% of the corpus population, 1.00 B each)
   reading reference frames that are immutable for the whole decode. Largest
   lever in the map, and **not an extent question**. Two cautions: do not reach
   for a hull (its 8 taps are 8 rows -> 27 KB at 4K -> the wide path, 2.65x), and
   the 4K vectors cannot see any of it.
2. `loopfilter.rs:710:14` — **demoted**, see above.
3. `ctx.rs:99:27` — 2,534,988/frame, **budget ZERO** (36 butt-adjacent writes).
   Count only, never extent.
4. **The CDEF quartet** — 1,563,240/frame combined, **0 collisions at `k<=64`**,
   which is 8-32x their current extent. MEDIUM: 64 B is close to `622:9`'s 80 B
   minimum.
5. 3.7 M/frame of `COARSENABLE-INF` context reads — note the **pairing**: at
   almost every one of those lines the sibling index on the same source line is
   `BLOCKED-0`. Two indices, one line, opposite verdicts.
6. **Close the instrument's own hole** — 1.2% `UNMEASURED`, one
   `probe_declare_rows` call per site. Cheapest item here.
7. **Re-run the map on what it has never seen** (item 1 was invisible until the
   corpus ran).

## Gates

| gate | result |
|---|---|
| corpus by name, **actual MD5 as the value**, vs `aarch64_md5_fixes_2026-08-07_final`, t=1 | **0 differences over 768 keys** — 766 PASS / 2 SKIP / 0 FAIL. Re-run at the LAST code-touching commit (`38f314b`), not inherited: two later commits touched the gate's test file, so the whole corpus was re-run and the two passes checked byte-identical rather than assumed. |
| the same at t=8 | **0 differences over the 755 shared keys** — 753 PASS / 2 SKIP / 0 FAIL |
| t=1 vs t=8 head-to-head | **0 differences** |
| Miri Stacked Borrows | 61 passed, 0 failed, 0 UB |
| Miri Tree Borrows | 61 passed, 0 failed, 0 UB (9 targets, `soundness` 993.1 s) |
| `__probe_wide` standing hazards (run although `crates/` is untouched) | baseline ok; both plants **FAILED**; restored byte-exact; ok again |
| `cargo test --lib` DEBUG / release | 75 passed / 75 passed |
| full `cargo test --release --workspace` | **81 passed, 0 failed**, 22 ignored across 41 targets (incl. `decode_md5_verify` 14 and `decode_permutations` 19) |
| `forbid(unsafe_code)` | proved ACTIVELY, anchored on `lib.rs:13` |
| warnings, release / debug / `+probe-sites` | 0 / 0 / 0 |
| `cargo fmt --all -- --check` | clean |
| clippy, both CI feature legs | clean |
| clippy reproduced on `x86_64-unknown-linux-gnu`, both legs | clean |
| clippy `c-ffi` on macOS | fails on the `EAGAIN` const-assert (35 vs 11) — **pre-existing**, identical failure A/B'd on the base |

**The 13-key gap at t=8 is not this branch's.** A whole-corpus
`md5_inventory --threads 8` cannot finish on `main` either: the first
`8-bit/film_grain` vector panics with `overlapping DisjointMut: &mut
_[0..147456]` (a whole-plane mutable guard taken by two workers), then
`src/thread_task.rs:534` unwraps a `None` and the process dies at 572 of 768
rows. The parent commit `be55376` reproduces it exactly — same 572 rows, same
569 PASS / 2 SKIP, same panic. #479's class, and the same exclusion #486 already
records for the probe. So the t=8 pass above is per-group with the two
film-grain groups excluded.

**No timing claim anywhere.** Nothing timed was run, so `measlock` was not needed
and not taken; all builds and tests were `nice -n 19`.

Based on `probe/bounds-map` (#486). Do not merge before it.


lilith:
**Closed. Split in two; both halves are MERGED to `main`, verified by SHA.**

| half | PR | verification |
|---|---|---|
| the record — `docs/BOUNDS_MAP.md` Part 2, `bounds_verdicts_2026-08-10.{tsv,meta}`, the two md5 gate TSVs, `scripts/perf/bounds_verdicts.py`, the 60 B correction | **#510** | `git merge-base --is-ancestor be1e8dd origin/main` → ok |
| the standing gate — `note_pic_extent` + `PIC_EXTENT_CEILINGS`, `tests/guard_extent_budget.rs`, CI `extent-gate` | **#511** | `git merge-base --is-ancestor 8c375b1 origin/main` → ok |

New branches rather than a rebase of this one: `probe/bounds-verdict` is checked out in another
agent's worktree (`~/work/zen/rav1d-safe--boundsverdict`) and AGENT_BRIEF §1 says never touch another
agent's branch. Also unstacked — this PR targeted `probe/bounds-map`, which merged as #486.

## The gate was re-validated against current `main`, and it holds

Its ceilings were fitted on `675a7ff`, twenty commits back, which is why it was split out rather than
riding the records. The CI leg (`--release --features probe-sites`, the corpus) is green and
non-vacuous:

```
extent budget: checks=588521423 checks_tile_threaded=588521423 max=3840 B
               max_under_tile_threading=3840 B whole_component_exempt=6313610
               fallback_ceiling=64 B frames=1734 undecodable_vectors=6
```

**588,521,423 reservations, all under tile threading, 1,734 frames** — up from this PR's 575,577,925,
so the population grew with `main` and every ceiling still holds. #505's `__lf_rect` does not collide
with rule 1 either: it registers through `dm().index_rect_as(..)`, not through the `slice_as` funnel
the gate hooks.

## The correction this PR made is now on `main`

`loopfilter.rs:710:14`'s nearest concurrent writer is its **own** write-back `loopfilter.rs:887:14`
at **60 B** over 1,176,771 co-live pairs — not `cdef_arm.rs:622:9` at 232 B, which is the *third*
nearest and was a per-pair figure quoted as the site's. `main` said 232 in **three** places:
`docs/OWNERSHIP_MODELS.md` §7d, `docs/AGENT_BRIEF.md` §6, and — which this PR missed —
`docs/BOUNDS_MAP.md`'s own "The counterparty has a name" bullet, where the figure was first
published. All three fixed; the erratum is appended to #486's `.meta`, not written over it.

## Two claims corrected in place, dated, nothing deleted

* **"#469 rectangle → BLOCKED, 18,196 collisions" is scope-corrected.** The ROWBAND counterfactual
  tests the **hull**, and the hull is blocked — that part is right. It is not what an *exact*
  rectangle registers: `eval_rect` reads `rect_ovl = 0` at every site, and **#505 built that record**
  — no `Deref`, per-row references derived from the buffer's own pointer, Miri-clean under both
  aliasing models, which also closes the second defect this section says the map cannot see — and
  measured a **t=8 win**, −1.0% to −1.8% wall on 5 of 6 multi-tile cells. Left as written with the
  scope named, so nobody carries "the rectangle is blocked" forward.
* **The candidate list's ranking key is superseded.** It ranks by registrations/frame; #505 measured
  that cost tracks **distinct shard lines visited**, not records filed — `LfBlock::fill` files 8.98
  records per 2.09 lines, is 31.7% of the population and **3.9-4.4%** of tracker CPU (2.42-2.71 ns
  marginal against a 19.71 ns average). A count cut at the densest site is the least valuable one
  available. The *verdicts* are unaffected; the ORDER is, and every row now says to re-price against
  `row_shards_mean`. Recorded alongside: `rows_mean / row_shards_mean` does not order the cells
  either, so **no published model currently explains the t=8 win**.

## One process note

This PR's `docs/AGENT_BRIEF.md` diff was **not** applied as a diff. Its base predates `ca26fd6`
("adopt the fullest version — the workspace copy had diverged and was stale"), so a mechanical rebase
would have silently reverted eleven §2 bullets and four §6 rows `main` has since gained — the
itx-census row, the `--load-ok` bullet, the running-shell-script bullet, the V-batch row, and more.
Only the substantive 60 B change was carried, by hand.

Nothing was re-measured on this box for either landing.

lilith:
**Correction to my close comment above.** I wrote that a mechanical rebase of this PR's `docs/AGENT_BRIEF.md` diff "would have silently reverted eleven §2 bullets and four §6 rows". Counted from the diff:

```
$ git diff 675a7ff pr489 -- docs/AGENT_BRIEF.md | grep '^-- \*\*' | wc -l
7      # §2 bullets deleted outright
$ git diff 675a7ff pr489 -- docs/AGENT_BRIEF.md | grep '^-| '     | wc -l
4      # §6 rows touched — but one of the four (the "Coarsening a guard extent"
       # row) is REPLACED, not removed, so net-deleted §6 rows = 3
```

**Correct figure: 7 §2 bullets deleted outright, plus a 3-line parenthetical inside an 8th (the `measlock` ownership note), plus 3 §6 rows net-deleted and 1 modified.** Not eleven. Written from recollection; corrected here and in #514, and #510's body is edited (its merged commit message cannot be).

The substantive point is unchanged and is still why the diff was applied by hand: this PR's base predates `ca26fd6`, so a mechanical rebase would have reverted real content. Only the 60 B correction was carried across.

PR #490, opened 2026-08-10
The hump is not a size law: ratio-to-dav1d vs pixels is R^2=0.001, and screen content is 2.1-3.2x (#455)

lilith:
Follow-on to `docs/SIZE_SWEEP.md` (branch `measure/size-sweep-intercept`, unmerged) and #455. Base SHA **`5606efe`** — the task named `0f6bf10`, which `main` had moved four docs-only commits past. Full record: **`docs/CONTENT_CLASS.md`**.

Worst news first.

## What is NOT done

* **No fix for the thing this round found.** It is a target redirect plus a profile, not a patch.
* **The one code change is 16bpc CDEF padding and it is NOT a measured win** — see the control arm below. Its sibling item (the itx scratch `memset`) was already fixed by #484, which also refuted the headline 16bpc-itx candidate by census; this branch adds nothing to 16bpc itx.
* Loop restoration is off in all 24 new vectors (`enable_restoration = 0`), so the campaign's structural blindness is inherited whole. No inter prediction, no 12bpc, no x86_64 or wasm run, no `asm`/`c-ffi`/`unchecked`, t=1 only for the content grid, one encoder (libaom 3.14.1).
* Screen content is two crops of two screenshots. That is not a content class; it is enough to show the axis moves the answer.
* **The box was never idle.** Two other agents held 100%-CPU jobs throughout; 0 of 540 content rows and 0 of 180 A/B rows are load-free. Paired within-round ratios only.
* Not run: t=2/t=4, 4:4:4 10bpc content cells, a 4K screen cell.

## 1. The census `SIZE_SWEEP.md` asked for — its inference was exactly right

`probe_tracker --features probe-sites`, t=1, counts only. Registrations per **megapixel** over its own ladder, 4:2:0 8bpc:

| | 64x36 | 256x144 | 512x288 | **1024x576** | **2048x1152** | 4K |
|---|---|---|---|---|---|---|
| regs/MP | 450,087 | 177,056 | 225,932 | **809,926** | **778,762** | 502,372 |
| normalised to 512x288 | 1.99 | 0.78 | 1.00 | **3.58** | **3.45** | 2.22 |
| its measured tracker ms/MP, normalised | — | — | 1.00 | 3.01 | 3.34 | 2.10 |

Every per-block site moves together (`ctx.rs:99` 3.50x, `decode.rs` 4.26x, `ipred_prepare` 3.60x, `recon.rs` 2.74x), so it is block **count**. CDEF's six sites are **exactly 0 at 512x288** and appear from 1024x576 up — a second instrument agreeing with `SIZE_SWEEP.md`'s "CDEF executes zero blocks there".

`ctx.rs:99:27` is **43.9% of every registration in the frame** at the hump cell, mean extent 3.3 bytes — twice as dominant as on `v4k_8tile` (22.2%, #488), so #488's un-reduced `t.l` half is worth ~2.7% of a 1024x576 frame here.

## 2. And it does not matter: the hump is not a size law

Block density is not a function of pixel count. 30 cells — 4 content classes x 4 qualities x 5 sizes:

```
ratio ~ log10(pixels)     : R^2 0.001
ratio ~ regs/pixel        : R^2 0.009
ratio ~ log10(dav1d ms/MP): R^2 0.578   slope -1.02
```

At **one** pixel count (1024x576) the ratio to dav1d spans **1.397 .. 3.164**, a 2.26x spread that contains the whole ladder's 1.12–1.49 range twice over:

| content class @1024x576 | q20 | q40 | q70 | q90 |
|---|---|---|---|---|
| photo, downscaled (the ladder's own content) | 1.824 | 1.616 | 1.432 | 1.444 |
| photo, native crop | 2.643 | 2.023 | 1.434 | 1.397 |
| **screen: wiki text** | **2.475** | **2.464** | **2.254** | **2.089** |
| **screen: macOS UI** | **3.164** | **3.115** | **2.897** | **2.573** |

The U with a hump at 0.6–2.4 MP is real and correctly measured; the *generalisation* to "0.6–2.4 MP is our bad zone" does not survive a second content class. n=9, plus an independent n=4 re-run that reproduces every cell within 0.12.

## 3. Screen content is 2.08–3.26x dav1d, and has never been measured

Every vector the campaign has used — `v256`, `v1024`, `v4k_*`, both 10bpc twins, all 24 size-ladder renditions — is a photo. `docs/AGENT_BRIEF.md` already required three content classes for source-informing sweeps.

## 4. Where it goes: the tracker is 38–48% of our excess in EVERY class

`sample`, same size and quality, only content varying, ~38k leaves each:

| bucket | photo | screen text | screen UI |
|---|---|---|---|
| entropy | **58.80%** | 11.37% | 13.22% |
| **borrow tracker** | 11.49% | **26.88%** | **30.58%** |
| kernels | 23.43% | 25.32% | 11.88% |
| `decode::read_pal_indices` | — | **20.93%** | **17.19%** |

In milliseconds, tracker share x measured frame time, against the same cells' excess over dav1d: **38.0% / 48.3% / 46.2% of the excess**, and dav1d has no borrow tracker at all. Its *share* triples on screen content not because it does more work — `simac` registers 497,582/frame against the photo's 477,714 — but because the frame is less than half as long. **Our tracker alone (2.35 ms) is 91% of dav1d's entire frame there.** That is the strongest case yet for #469 / #474 / #481, and photo-only measurement could not see it behind a 59%-entropy wall.

Also: `read_pal_indices` is a per-pixel palette symbol loop the bucketer files under "other" (so entropy's true screen share is ~30%, not 11%), and a key-frame-only AVIF is running `refmvs::splat_mv` + `add_spatial_candidate` — intra block copy through the inter path.

## 5. The code change, and why it is not a win

`padding_8bpc` copies rows through `widen_n::<N>` (compile-time length, inline); `padding_16bpc` needed no widening so it used `copy_from_slice`, whose runtime length lowers to a `_platform_memmove` call for 4–12 `u16`s, up to `h + 4` times per CDEF block. `SIZE_SWEEP.md` Q2 named it (381 of 549 memmove samples from `cdef_filter_block_16bpc_inner`, zero from 8bpc CDEF). `copy_row_u16` is `widen_row`'s twin over the same length set. Byte-identical by construction.

**Mechanism verified:** memmove inclusive **636 → 201 samples (−66%)**, and `cdef_filter_block_16bpc_inner` disappears from the caller list entirely.

**Wall clock, n=9 paired:** the only defensible cell is `L64x36_420_10b` **0.9782, 9 of 9 rounds faster, p=0.004**. Everything else is inside the instrument — `L256x144_420_10b` reads **+0.78% (1/9, p=0.039)**, and an **8bpc control cell the change cannot execute reads +0.94% (3/9)**. Two binaries differing only inside `if BD::BPC == BPC16` still have different code layout, and that alone is ~1% here. No cell is band-disjoint.

Kept for consistency with `padding_8bpc`, not for the number; revert `be289f4` alone if you would rather not carry it.

## Gates

* **Corpus by name with the md5 as the value:** `--threads 1` **766 PASS / 768 keys / 0 differing, SETDIFF CLEAN**; `--threads 8` with `8-bit/film_grain` + `10-bit/film_grain` dropped from **both** sides (#479) **753 / 755 / 0, CLEAN**.
* **New vectors bit-identical to dav1d 1.5.4 before any timing: 24/24 MATCH.**
* **Teeth planted, confirmed failing, restored byte-exact:** `copy_n` writing `a[0]` into its last lane → `10-bit/data` **71 MISMATCH of 71** while `8-bit/data` stays **358 PASS** (proves the path is live on 71 corpus vectors *and* 16bpc-only). sha256 identical before/after, `git diff --exit-code` clean.
* **`#![forbid(unsafe_code)]` proved ACTIVELY** from `src/ablate.rs`: error anchors on **`lib.rs:13:12`** and overrules `#[allow(unsafe_code)]`. Restored byte-exact.
* **Standing hazards under `--features __probe_wide`** (run although `crates/` is byte-identical to base, sha `d4e03d4a…`): control 6/6 pass; deleting the in-lock `state` re-read FAILS `wide_exclusion`; `active()` cut to one shard FAILS it. Restored byte-exact.
* **Miri, both models, `-p rav1d-disjoint-mut --no-fail-fast`:** Stacked Borrows **9 targets, 61 passed, 0 failed, rc=0, no UB**; Tree Borrows in flight at hand-off (29 passed, 0 failed so far).
* `mt_stress` (1/2/4/8/16 internally) **ok**; `multi_decoder_pressure.sh` 12 concurrent decoders x 3 iters over 10bpc + screen vectors, **all md5s match the serial reference**.
* `cargo test --lib` in the **debug** profile: **75 passed, 0 failed**.
* **x86 clippy reproduced locally** (`--target x86_64-apple-darwin --all-targets --keep-going -D warnings`): 11 sites head, 11 base, **set-diff empty both directions**, none in the changed file.

## Two harness traps, added to `docs/AGENT_BRIEF.md`

* **`kill`ing a `measlock` released the lock and kept running the payload.** A bash trap handler returns unless it exits, so a TERM'd holder deleted its own lock, sat out the 20-minute politeness wait unlocked, then started a second timed sweep on top of another — truncating the first run's TSV. Fixed in `~/bin/measlock` (signal handlers terminate; payload backgrounded so `wait` is interruptible), verified against the old version, which survives the same TERM.
* **A same-code control arm is the only way to know a sub-2% A/B's floor** (§5).

## Ranked next step

1. Profile-and-fix screen content — 2.1–3.2x, never measured, §4 names the leaves.
2. `ctx.rs:99`'s `t.l` half — 22% of all registrations here, borrowck as the proof, no `unsafe`, ~2.7% of a 1024x576 frame; attributed by #488, not reduced.
3. Stop selecting targets by rendition size (R² = 0.001). A gap number needs its content class and quality named next to it.

Do not merge without a maintainer read of §5's revert rationale.


lilith:
**Closed. Split in two and both halves are open/landed as separate PRs, because this PR's own body asked for that:** *"Do not merge without a maintainer read of §5's revert rationale."*

| half | where it went | state |
|---|---|---|
| the record — `docs/CONTENT_CLASS.md`, 12 benchmark files, 3 harness scripts, the 2 AGENT_BRIEF §2 bullets | **#508** | **MERGED**, verified `git merge-base --is-ancestor 7390dde origin/main` |
| the code — `src/safe_simd/cdef_arm.rs` `copy_row_u16` | **#509** | **OPEN**, for the maintainer read this PR asked for |

A new branch rather than a rebase of this one, because `perf/size-hump-itx16` is checked out in
another agent's worktree (`~/work/zen/rav1d-safe--sizehump`) and AGENT_BRIEF §1 says never touch
another agent's branch. Nothing here was force-pushed.

## The finding landed unchanged, because nothing merged since contradicts it

R² = 0.001 for ratio-to-dav1d on `log10(pixels)` over 30 cells; a 1.397..3.164 spread at **one**
pixel count; screen content 2.08-3.26x against the photo ladder's 1.12-1.49x; the tracker at 38-48%
of our excess in every class and 91% of dav1d's whole frame on screen UI. §8 item 3 — *"stop
selecting targets by rendition size"* — is arguably the most load-bearing sentence in the file and
went in verbatim.

## Two things were corrected in place, dated, nothing deleted

* **§8's ranked item 2** (`ctx.rs:99`'s `t.l` half, *"attributed by #488, not reduced"*) is **DONE**:
  #492 (write side, −11.8% wall at t=8 on screen content), #495 (the read side) and #502 (the LEFT
  neighbour's reads leave the tracker, 0.9842 geomean) all merged. Item 1 is no longer un-started
  either — `text_q20`/`ui_q20` have been standing cells since #502.
* **§5's control-cell inference** — *"~1% of this A/B is code LAYOUT, not the change"* — is
  **CONFIRMED and generalised by #506**, which built the instrument this PR was reasoning without:
  `--features __pad_text` costs **+1.1% to +1.6%** at t=1 from provably-dead `#[used]` text with
  **0 symbols resized**, while a byte-identical copy reads 1.0006. This PR inferred that a day
  earlier from a single control cell and got it right; the floor generalises to **±1.5% on any t=1
  arm**. The AGENT_BRIEF bullet carries the cross-link so the §2 (cell control) and §6 (binary
  control) halves read as complementary rather than competing.

## One thing #508 could not fix

`docs/SIZE_SWEEP.md`, which `CONTENT_CLASS.md` cites **eleven times**, is still not on `main` — it
lives only on the unmerged branch `measure/size-sweep-intercept` (`30a7294`). The relative markdown
link at the top of the file would have 404'd, so it is now an explicit "not on main, here is where
it is" pointer. Landing `SIZE_SWEEP.md` is a separate decision.

## On the code half (#509)

Kept whole and lifted verbatim from `be289f4`. Its mechanism is verified independently of the clock
(memmove inclusive 636 → 201 samples, −66%, `cdef_filter_block_16bpc_inner` gone from the caller
list) and its verdict is unchanged: **not a measured win** — only `L64x36_420_10b` at 0.9782 (9/9,
p=0.004) is defensible, `L256x144_420_10b` reads a nominally-significant *regression* at +0.78%
(1/9, p=0.039), and the 8bpc control the change cannot execute reads +0.94%. Every `cdef16_*` record
file landed in #508 **regardless**, because it is the evidence for the *no* as much as for the
mechanism. Ship it for consistency with `padding_8bpc` or close it; both are defensible.

Nothing was re-measured for either landing — another agent holds the box for timed work.

PR #491, opened 2026-08-10
fix(#479): film grain could not decode above 1 thread; and LR measured — LR is fine, the tracker is the pathology

lilith:
## What is INCOMPLETE or BROKEN first

* **The LR NEON kernel port asked for in the task was NOT done, deliberately, and should not be.** Measured: loop restoration is **1.88x** of dav1d's LR at 4K 8bpc and **2.25x** at 1080p 10bpc while the rest of the decoder is **3.9-6.8x**. It is 9.3% of our wall at the worst cell. See §2 — the task's premise that it "runs scalar while everything around it is vectorised" is false.
* **`intrabc` / screen content: NOT measured.** dav1d-test-data has no screen-content group, so a vector must be encoded. It is structurally suspect: `mc_put_dispatch` bails to the scalar `PicOffset` path whenever `dst.data.ref_eq(src.data)` (`mc_arm.rs:5648`, `mc.rs:12110`, `:12832`), so **every intrabc block runs scalar mc by construction**. Named, unpriced.
* **Monochrome I400 — which is AVIF alpha's layout — has exactly ONE corpus vector** (`8-bit/data/00001105`), and dav1d decodes it below the 1 ms timer floor, so no honest ratio exists for it. Any I400 defect is near-invisible to the gate.
* **12 bpc and 4:2:2 ratios are granularity-bounded.** The only 12-bit geometry in the corpus is 160x90x10 frames; with a 1 ms clock over an 8-frame delta the resolution is ±0.125 ms/frame. Read `5.5x`/`11.0x` as an order of magnitude, not two significant figures.
* **The tracker fix implied by §3 is not attempted here.** This PR measures and attributes it; it does not touch it.
* **Absolute ms in every table below are inflated** — a sibling agent held one core with `miri` for the whole sweep, so all rows are `foreign_max=1` under `measlock --load-ok` / `ALLOW_LOAD=1`. Arms interleave back-to-back with a rotating order inside each cell, n=9, 0 discarded, so **paired ratios are sound and absolutes are not**.
* **The negative-stride fallthrough added to the film-grain dispatch has no test vector.** No corpus stream has a negative stride; only a caller-supplied `Rav1dPicAllocator` under `c-ffi` can produce one. It replaces a guaranteed index-out-of-bounds panic (present on `main` too, at any thread count) with the correct scalar path, unexercised.

---

## 1. #479 — film grain could not be decoded above one thread at all. Fixed.

`rav1d_apply_grain_row` is claimed by N workers that each `fetch_add` a **different** `FG_BLOCK_SIZE` row band off `TaskThreadData::delayed_fg_progress[0]` (`src/thread_task.rs`, `TaskType::FgApply`). All four film-grain guards in `src/safe_simd/filmgrain_arm.rs` reserved the **whole picture component** per band, so the workers collided on their first band each:

```
thread 'rav1d-worker-4' panicked at include/dav1d/picture.rs:736:19:
	overlapping DisjointMut:
 current: &mut _[0..147456] on ThreadId(6) at include/dav1d/picture.rs:736:19
existing: &mut _[0..147456] at include/dav1d/picture.rs:736:19
...
thread 'main' panicked at src/thread_task.rs:534:56:
called `Option::unwrap()` on a `None` value
```

**Consequence: 13 of 768 corpus vectors could not be decoded at any thread count above 1**, nondeterministically, and every corpus run of this campaign passed `--skip-group film_grain` to work around it.

Each guard is narrowed to the band the call actually touches:

* dst/src — the strided hull `(bh - 1) * stride + pw` from the band's first pixel.
* `fguv`'s luma read — rows `y << sy` for `y in 0..bh`, so the last luma row is `(bh-1) << sy` and **not** `bh << sy`, by columns up to `((pw-1) << sx) + sx` (because `noise_uv` reads `luma_row[(x << sx) + 1]` when `is_sx`).

Bands are provably disjoint: consecutive bands start `FG_BLOCK_SIZE * stride` apart (`>> ss_y` in a subsampled plane, matching the band height there), a band spans at most `(FG_BLOCK_SIZE - 1) * stride + pw`, and `pw <= stride` always.

This is a **narrowing**, the direction `docs/OWNERSHIP_MODELS.md` says is always available: the three refuted schemes all *widened* a reservation — §3 partial keying, §6 the exact-record/wide-reference that was outright UB, §7d the contiguous filter band that swallowed a concurrent narrow write. Record and reference here are the same hull, which §6 makes non-negotiable.

**The gate is removing the workaround, and it is removed:**

```
TOTAL threads=8 pass=766 mismatch=0 error=0 skip=2 skipped_groups=-
TOTAL threads=1 pass=766 mismatch=0 error=0 skip=2 skipped_groups=-
```

Set-diffed **by name with the actual MD5 as the value** against `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst` at both thread counts: `768 rows both sides / 0 only-in-baseline / 0 only-in-head / 0 differing`. Re-run after the last code-touching commit.

`tests/filmgrain_threads.rs` is the regression gate — every film-grain vector x {1,2,4,8} threads, 3 reps above t=1, plus two liveness tests (grain must actually change pixels; some vector must span more than one row band, else two workers can never contend). **Teeth proven by mutation**: reverting one guard to `full_guard_mut` aborts it with the exact signature above; restored, sha256-verified identical, green.

Why the existing gate could not see this: `decode_md5_verify`'s film-grain tests use `Settings::default()`, which is `threads: 1`.

## 2. The other three whole-plane-guard sites, audited (OWNERSHIP_MODELS §7e)

| site | verdict |
|---|---|
| `loopfilter.rs:140,182` + `looprestoration.rs:212,258` | **unsound under concurrency, NOT a shipping defect** — all four are inside `#[cfg(feature = "__simd_test")]`. Measured: at `--threads 8 --group 8-bit/data` they give **313 errors in 358 vectors** (`& _[0..983040]` vs a concurrent `&mut _[163840..163968]`). **Narrowing cannot fix it** — the harness semantically needs the whole plane (save → restore → run scalar → write SIMD back), so it would clobber other workers even with tracking off. Already quarantined: the only threaded test in `decode_md5_committed.rs` is `#[cfg(not(feature = "__simd_test"))]` and CI's `__simd_test` step runs only that file (re-verified passing). `examples/md5_inventory` now **fails loud** on `--threads > 1` under `__simd_test`. |
| `safe_simd/mc.rs` x6 + `mc_arm.rs` x4 | **safe.** Immutable, so no mutual conflict; the only question is whether a mutable borrow of the same allocation can be live, and three mechanisms say no — refs are complete because `n_fc` is pinned to 1 without `unchecked` (`lib.rs:127`), and `unchecked` makes picture buffers `dangerously_unchecked` (untracked); intrabc is refused by `ref_eq` *before* the guard; `mct_prep_dispatch` writes a scratch `&mut [i16]` with `src` only ever `&f.refp[..]` (`Temp` is compound inter + warp only — `recon.rs:2927`, `:2941`, `:3025`, `:3039`; intrabc uses `NonTemp` at `:2876`/`:2893`, OBMC writes into `lap` at `:1945`/`:1997`). |

## 3. Loop restoration on real content: LR is fine, the TRACKER is the pathology

Full record: `benchmarks/lr_gap_2026-08-10.meta`. Host Apple M4 Pro (8P+4E), macOS 26.5.2, dav1d 1.5.4 in the same interleaved sweep at both `--framedelay 1` and its shipping default, whole-process two-point wall fit, per-cell frame counts, n=9.

**Two premises checked before anything else, both false on `main` @ 312dcc3:**

* `looprestoration_arm.rs` is **not** "zero aarch64 intrinsics". 1,812 lines, **61 distinct NEON intrinsics over 249 call sites** — more sites than `ipred_arm.rs` (88), `cdef_arm.rs` (44) or `filmgrain_arm.rs` (30). `perf/lr-neon-tier` (edb55a1) is on main.
* LR *had* been measured, 2026-08-08, `verify_compose3_2026-08-08.meta` §5. Cited, not rediscovered; everything here is a fresh independent sweep on a wider ladder.

Confirmed by re-running it: **LR active in 696 of 768** vectors. Also found that six of `md5_inventory --activity`'s nine columns are **uninstrumented** (only itx/cdef/looprestoration call `ablate::note()`), so `loopfilter 0` reads as "never ran" — the harness now says so out loud.

**LR is differenced on BOTH decoders with one instrument** (`RAV1D_INLOOP` / `--inloopfilters`). Proof the two arms mean the same thing: at `all`, `norestoration` and `none` our md5 is **byte-identical to dav1d 1.5.4's**.

| cell | bpc | t | ours | noLR | our LR | LR% | dav1d | noLR | d1d LR | **LR ratio** | REST ratio |
|---|---|---|---|---|---|---|---|---|---|---|---|
| 3840x2160 | 8 | 1 | 160.50 | 145.50 | 15.00 | 9.3% | 40.50 | 32.50 | 8.00 | **1.88x** | 4.48x |
| 1920x1080 | 10 | 1 | 7.87 | 7.27 | 0.60 | 7.6% | 2.13 | 1.87 | 0.27 | **2.25x** | 3.89x |
| 640x360 | 8 | 1 | 3.08 | 3.00 | 0.08 | 2.4% | 0.47 | 0.45 | 0.02 | 3.00x | 6.67x |
| 352x288 I444 | 8 | 1 | 5.20 | 5.09 | 0.11 | 2.2% | 0.80 | 0.75 | 0.05 | 2.17x | 6.80x |
| 240x100 | 8 | 1 | 0.69 | 0.68 | 0.01 | 1.3% | 0.15 | 0.15 | 0.00 | 4.00x | 4.66x |

**LR is the smallest ratio in the table.** Zeroing it entirely at 4K t=1 moves 3.96x -> 3.59x. Do not port an LR kernel on this evidence.

**What is pathological — the real-content ratio, and thread scaling that INVERTS:**

| cell | bpc | t=1 ratio | t=8 ratio | ours t1->t8 | dav1d_fd1 t1->t8 |
|---|---|---|---|---|---|
| 3840x2160 | 8 | 4.00x | 8.20x | **1.15x SLOWER** | 0.56x faster |
| 1920x1080 | 10 | 3.67x | **46.17x** | **3.54x SLOWER** | 0.30x faster |
| 640x360 | 8 | 6.53x | 7.78x | 1.13x SLOWER | 0.95x |
| 352x288 I444 | 8 | 6.49x | 9.61x | 1.27x SLOWER | 0.87x |
| 240x100 | 8 | 4.65x | 5.68x | 1.38x SLOWER | 1.13x |

Every LR-active cell is slower at 8 threads than at 1, ranges disjoint at 5 of 6, **and it happens with LR off too** (1.15x-3.75x). ~1.4x is met at no cell on this content.

**Cause isolated** with `--features probe-untracked` (bit-identical output verified first):

| cell | bpc | t | ours | untracked | tracker share | ours/dav1d | untracked/dav1d |
| 3840x2160 | 8 | 1 | 160.00 | 101.00 | 36.9% | 4.00x | 2.52x |
| 3840x2160 | 8 | 8 | 184.50 | 70.50 | 61.8% | 8.20x | 3.13x |
| 1920x1080 | 10 | 4 | 16.10 | 2.27 | 85.9% | 24.15x | 3.40x |
| 1920x1080 | 10 | 8 | 27.70 | 1.93 | **93.0%** | 46.17x | **3.22x** |
| 640x360 | 8 | 8 | 3.50 | 2.10 | 40.0% | 7.78x | 4.67x |
| 352x288 I444 | 8 | 8 | 6.60 | 3.37 | 48.9% | 9.61x | 4.90x |
| 240x100 | 8 | 8 | 0.96 | 0.54 | 43.0% | 5.68x | 3.23x |

All pairs disjoint. Thread scaling at 1920x1080/10b: tracked 1.13x / 2.06x / 3.54x at t=2/4/8; **untracked 0.63x / 0.37x / 0.32x**. The tracker converts a 3.1x speedup into a 3.5x slowdown.

`sample` self-time on `10-bit/issues/318_tx_4x4` (in-process ms/frame agreed with the two-point fit to 3.1-3.9%):

| t=1 (19,061 leaves) | t=8 (163,910 leaves) |
|---|---|
| 22.2% `recon::mc<BitDepth16>` | **39.5% `TinyLock::lock_slow`** |
| 11.3% `_platform_memmove` | 39.3% `__psynch_cvwait` |
| 6.1% + 6.1% `BorrowTracker::add` | 5.3% + 3.8% `BorrowTracker::add` |
| 2.3% `lr_apply::lr_stripe` | 0.8% `recon::mc` <- the pixel work |
| ~5.8% all LR, ~13.2% all tracker | ~55.3% all tracker |

**Note for the brief's "`TinyLock` backoff/yield — null, measured twice":** both nulls were taken on the 4K multi-tile grid, where `lock_slow` is not the top symbol. This is the contradicting profile.

## 4. Cheap hunt for other pathological features

| feature | t=1 | t=8 | verdict |
|---|---|---|---|
| film grain, t>1 | — | — | **WAS BROKEN** (13/768 undecodable above 1 thread); fixed here, 766/766 at t=8 with no skip |
| 12 bpc | 5.5x | 11.0x | runs; all 47 12-bit vectors PASS at t=8; granularity-bounded (one 160x90x10 geometry exists) |
| 4:4:4 at t>1 | 6.49x | 9.61x | runs; 68 I444 vectors PASS at t=8 |
| 4:2:2 | 8.0x | 4.0x | runs; 30 I422 vectors PASS; granularity-bounded |
| monochrome I400 (= alpha) | n/a | n/a | runs; **one** corpus vector, dav1d below the timer floor; 3.00x vs dav1d default |
| resize / superres | 3.67x | 4.83x | runs; 9.67x vs dav1d default at t=8 |
| all-intra | **1.48x** | **1.48x** | **best cell measured anywhere** (1.31x untracked) — and it is the AVIF-still shape. 4.67x vs dav1d default at t=8 |
| intrabc / screen content | — | — | **NOT measured**; forced onto the scalar mc path by the `ref_eq` bail |
| AVIF grid / alpha items | — | — | not a decoder axis (zenavif-parse container features; each item is an ordinary AV1 stream) |

## 5. Gates

* **`#![forbid(unsafe_code)]` proved ACTIVE by mutation, twice**, in always-compiled files: planting `unsafe { 1u8 }` in `lib.rs` fails with `error: usage of an unsafe block` anchored at **`lib.rs:13:12  forbid(unsafe_code)`**; planting one in `src/safe_simd/filmgrain_arm.rs` (the file this PR changes) fails anchored at `filmgrain_arm.rs:8:48`. Both restored, sha256-verified, `git diff --exit-code` clean.
* **Corpus set-diff BY NAME with the MD5 as the value**, re-run after the last code-touching commit, at t=1 **and** t=8: CLEAN both, 766 PASS / 2 SKIP, 0 differing. No `--skip-group`.
* **Standing hazards under `--features __probe_wide`** (run even though `crates/` is untouched — `git diff 312dcc3 -- crates/` empty, `tracker_shard.rs` sha256 `d4e03d4a…`, the value #482 recorded): 9 targets, all pass — soundness 25, lib 26, shard_liveness 5, guard_move_release 2, narrow_release 1, wide_exclusion 1, doctest 1. Crate tree restored byte-exact (23 files sha256-verified) and `git diff --exit-code` clean.
* **Miri, both memory models, each target in isolation** (`--no-fail-fast`): see the results comment.
* **`cargo test --lib`**: release 75 passed / 8 ignored; **debug** 75 passed / 8 ignored.
* **`mt_stress`** at threads 1/2/4/8/16 x 5 trials: ok.
* **`multi_decoder_pressure.sh`**: 16 concurrent decoder processes then 3 rounds at 24, thread counts cycling 1/2/4/8/16 over five vectors, hard wedge deadline — **4/4 rounds PASS**, every md5 equal to the serial reference.
* Integration suite green: `decode_md5_verify` 14, `decode_cpu_levels` 6, `decode_md5_committed` 3, `filmgrain_threads` 3, `safe_simd_crashes` 15, `tile_threading_parity` 3, `cancellation` 6, `panic_safety_test` 4, `managed_api_test` 4, `thread_cleanup_test` 2, `integration_decode` 1.
* `cargo fmt --check` clean.

## 6. Harness changes, all additive

* `scripts/perf/verify_gap_ivf.sh` — `ARMS` entries take an `@<inloopfilters>` suffix, mapping to `RAV1D_INLOOP` on a rav1d arm and `--inloopfilters` on a dav1d one, so a filter's cost is differenced on both decoders with one instrument.
* `examples/bench_ivf_limit` + `examples/profile_ivf` — `RAV1D_INLOOP` (dav1d's spelling, deliberately).
* `examples/profile_ivf` — `RAV1D_THREADS`, default 1 so older records stay comparable. Without it the t=8 slowdown could not be profiled at all: `bench_ivf_limit` is thread-capable but exits before `sample` can attach.
* `examples/md5_inventory` — `w`/`h` columns (a family census without sizes cannot separate per-pixel cost from a per-call intercept), a loud note naming the three instrumented families, and a hard failure on `--threads > 1` under `__simd_test`.


lilith:
## Gate results — Miri (both models) and the follow-up commits

**Miri, both memory models, each target in isolation** (`cargo +nightly miri test -p rav1d-disjoint-mut --no-fail-fast`), run even though `crates/` is untouched by this branch (`git diff 312dcc3 -- crates/` empty; `tracker_shard.rs` sha256 `d4e03d4a70183660cde4ef18cde777d5ef29530501c5a0e029a524e9c423176d`, the value #482 recorded):

| model | targets | result | UB |
|---|---|---|---|
| Stacked Borrows (`MIRIFLAGS=""`) | 9 | **61 passed, 0 failed**, `rc=0` (lib 26, soundness 25, shard_liveness 5 @ 992.7 s, guard_move_release 2, narrow_release 1, wide_exclusion 1, doctest 1) | none |
| Tree Borrows (`-Zmiri-tree-borrows`) | 9 | **61 passed, 0 failed**, `rc=0` (shard_liveness 993.1 s) | none |

Both reproduce #482's recorded 61/0.

**`decode_permutations` is GREEN, and the brief's trap #5 is now stale.** 19 passed / 0 failed in 776.93 s at `--test-threads=1`. `verify_compose3_2026-08-08.meta` §9 recorded 3 passed / 16 FAILED on main's lineage, caused by `Arm64::summon().unwrap()` sites in `mc_arm.rs` / `filmgrain_arm.rs`; those have since been converted to gates on main (`filmgrain_arm.rs` now carries an explicit `// GATE, don't assert` comment and a scalar twin). Not attributable to this branch — the harness runs at `Settings::default()`, i.e. one thread, so the #479 fix cannot be what changed it. Recording it so the next session does not treat a red `decode_permutations` as normal.

## Two corrections to my own numbers, both toward less confidence

1. **The LR *ratio* only resolves at 4K** (`ff85ac3`). It is a quotient of two small differences, so it must be banded. Differencing per round and taking the median/range of the nine quotients:

   | cell | our LR ms/frame | dav1d LR ms/frame | paired LR ratio | n |
   |---|---|---|---|---|
   | 3840x2160 8b | 15.50 [13.00..19.00] | 7.50 [7.00..8.50] | **1.94x [1.75..2.57]** | 9 |
   | 1920x1080 10b | 0.60 [0.43..0.77] | 0.27 [0.17..0.30] | 2.50x [1.63..4.20] | 9 |
   | 640x360 8b | 0.12 [0.05..0.18] | 0.02 [0.00..0.05] | 3.00x [2.00..6.00] | 7 |
   | 352x288 I444 | 0.10 [0.03..0.18] | 0.05 [0.03..0.05] | 2.33x [0.50..7.00] | 9 |

   Only 4K 8bpc is resolvable. Below it LR's own cost sits on the 1 ms timer floor and the 352x288 band **crosses 1.0**, i.e. means nothing. The `1.88x` / `2.25x` / `3.00x` / `2.17x` points in the PR body are medians of arm medians; the per-round bands above supersede them.

   **The "do not port LR" conclusion does not rest on the ratio.** It rests on the SHARE, which is measured directly and whose ranges *are* disjoint: ours 4K t=1 [159.50..164.50] vs noLR [144.50..148.00] = 9.3%, smaller cells bounded above by 2.4%. A kernel worth 9.3% of the worst cell cannot move a 4.0x ratio to 1.4x even if it became infinitely fast.

2. **The single band length rests on an equal-stride precondition that the old guards did not need** (`2793404`). `fgy_inner_*` / `fguv_inner_*` take a SINGLE `stride`, so equal `dst`/`src` strides were already a kernel precondition and `rav1d_prep_grain` asserts it — but the whole-plane guards took each component's *own* length, so they did not depend on it. Now stated with `debug_assert_eq!` on both dispatches rather than left implicit. Corpus re-gated 766/766 at t=1 and t=8 (set-diff by name, CLEAN) and `filmgrain_threads` green in DEBUG, where the assert is live, as well as release.

## Film-grain coverage, named (`7ddd132`)

The corpus has 13 film-grain vectors and that is the whole of the coverage:

```
 8-bit  I420  64x63 (odd height), 352x288
 8-bit  I422  64x63 (odd height), 63x64 (odd WIDTH)
 8-bit  I444  64x64 x3
10-bit  I420  352x288
10-bit  I422  64x63 (odd height), 63x64 (odd WIDTH)
10-bit  I444  64x64 x3
```

All three layouts and both depths are exercised, and the odd-width `ss_x` path — where `rav1d_apply_grain_row` writes two padding pixels into the INPUT luma plane while other workers read it, which is the *second* instance of #479 that the narrowed luma guard fixes — is covered by the two `422_oddwidth` vectors. **Missing: 4:2:0 at odd width** (nothing exercises `ss_x` and `ss_y` together with the padding write), **12-bit film grain** (no vector at all), and **any grain frame above 352x288** — so the narrowed extents are only ever checked at small strides, with at most 9 row bands.

Issue #458, opened 2026-08-08
x86-64 single-thread decode +59% regression from verify/compose-2 tracker changes (aarch64 unaffected)


lilith:
## Summary

x86-64 **single-thread** decode regressed ~+59% on `main` with the verify/compose-2 merge (#454): v4k_8tile 8bpc t=1 went 220.6 → 350.8 ms/frame on an idle Ultra 7 265K. aarch64 did not regress — the compose-3 record's own M4 numbers show t=1 slightly *improving* (comp/base 0.994). The MT scaling win is real on x64 too (t1→t8 0.36× → 3.66×), but it now sits on top of a much slower t=1.

## Bisect (i265, idle, interleaved 3 rounds × 3 reps × 2 iters, `bench_ab_decode v4k_8tile.avif 1`)

| build | median ms/f | range |
|---|---|---|
| pre-merge main `a6a7e232` | **220.6** | 215.6..227.1 |
| verify/compose-2 merge `490ed78a` | **350.8** | 347.0..357.3 |
| verify/compose-3 merge `79343e64` | 352.7 | 349.4..359.4 |
| main `fc120a8b` | 352.6 | 346.7..360.0 |

→ introduced entirely within **verify/compose-2**; compose-3 is neutral. 10-bit shows the same shape (275 → 412 ms/f, +50%).

Code commits in that window (docs/bench/fmt excluded): `1f09769` (lock-free release + re-opened shift ladder), `c003d2f` (block shift = function of the buffer, not a constant), `94f1bdb` (held-row-guards, default off — presumed inert), `fd5239f` (adaptive shift by default once parallelism declared), `2831954`/`27182197` (lr-itx-neon — aarch64-gated, presumed inert). Fine-bisect over these is running; will update with the exact commit.

Note: at t=1 the adaptive shift should be gated off (`set_parallelism(1)`, and the tile gate `c8663e7c` post-dates compose-2), so the naive expectation was "no t=1 change" — the M4's t=1 ladder column was "flat-to-slightly-adverse". On x64 it is +59%, which suggests either the serial-path shift/shard selection changed (`c003d2f`), or the lock-free release (`1f09769`) costs more on x86 (lock-prefixed RMW per release) than on M4, or an interaction neither record measured on x86.

## Context

- Product impact: threads=1 is `managed::Settings::default()` — this is the default ST decode path.
[...]

PR #493, opened 2026-08-10
verify(x64): first x86_64 execution pass — 766/766 at t=1, four t=8 aborts, A6+A1 refuted

lilith:
Verifies `docs/X64_APPLICABILITY.md`, which had never been tested. Base `main` @ `5e9975f`
(the task named `312dcc3`; `main` moved mid-session — every number here is against `5e9975f`).

Full record: `benchmarks/x64_verify_2026-08-10.meta` + the three artifact files. Narrative and
per-row verdicts: new **section F** of `docs/X64_APPLICABILITY.md`.

## The commissioning premise was FALSE — read this first

The task said "essentially nothing has been measured on x86_64; x86 has only ever been
compile-checked". **It has been.** `benchmarks/x64_i265_{gap,postmerge}_2026-08-08.meta` measured
this decoder on an **Intel Core Ultra 7 265K** (Ubuntu, idle, dedicated, dav1d built on-box,
interleaved 3x3) two days before section A was written — and `docs/X64_APPLICABILITY.md` cites
neither file. New **section F9** pulls the standing numbers forward (restricted to the arms
`x64_i265_CORRECTION_2026-08-08.md` leaves valid; `audit445`/`attrib` are VOID), **cited, not
re-measured by me**:

* x86 single-thread is **narrower** than aarch64 vs dav1d: **1.51x / 1.99x** (8/10bpc) against the
  same-week aarch64 1.62x / 2.01x.
* **x86 tile threading ANTI-SCALES**: 220.1 / 428.7 / 590.6 / 622.2 ms/f at t=1/2/4/8 — t=8 is
  **2.8x slower than t=1** while dav1d scales 4.46x on the same box. So the gap goes **1.5x -> 19.0-20.8x**,
  where aarch64 reads 3.44x. **That is the materially-different x86 shape the task asked for, and
  it dwarfs every row in section A.**
* The campaign's **sharded-tracker chain costs x86 +59% at t=1** (issue #458 + the CORRECTION's
  re-bisect), aarch64 flat over the same merge. The base I verified (`5e9975f`) is post-merge, so
  that regression is in it.

Consequences: A1's "applies, re-fit the ladder" is optimistic in the wrong direction, and **A2's
"fully applicable" and A5's batch factor have no valid x86 datum at all** — the only arm that
composed them is VOID.

## Incomplete / not measured by THIS pass

TCG has no cache model, no store buffer and no PMU, so **no wall-clock number came out of this
pass and none can**:

1. **The gap to dav1d on x86_64 — nothing new from me** (see F9 for what is already on record).
2. **A1's block-shift ladder — NOT MEASURED** (same reason).
3. **#482's seam cost (0.3-1.3% on aarch64) — NOT MEASURED.** The static half *is* checked: no
   out-of-line `ReconSrc::slice` symbol in either arch's release binary, so #483's symptom does not
   appear on x86. Liveness of the owned path on x86 is also unproven.
4. **The AVX-512 tier is entirely unexecuted** — 1,754 `_mm512_*` sites. Everything below is about
   the **AVX2 tier only**.
5. One unexplained x86 stall (`__do_sys_pause`, 0% CPU, single thread, and no `pause` call site in
   the binary). Recorded, not attributed.

**What would close 1-4: one `ubuntu-latest` CI leg** running `md5_inventory --threads 8` would have
caught the defect below; the perf rows need a bare-metal x86 host.

## Correctness: x86_64 is clean at t=1, and section B is CONFIRMED

| arch | threads | PASS | MISMATCH | ERROR | SKIP |
|---|---|---|---|---|---|
| x86_64 | 1 | **766** | **0** | **0** | 2 |
| x86_64 | 8 | 762 | 0 | **4** | 2 |
| aarch64 | 1 | 766 | 0 | 0 | 2 |
| aarch64 | 8 / 16 / 32 | 753 (film_grain skipped, #479) | 0 | 0 | 2 |

Zero MISMATCH at both thread counts ⇒ all eight kernel defects in section B were aarch64-only, as
claimed. Proven to be about the vector kernels rather than a scalar fallback by a new tier census
(`--features __probe_x86tier`): **384,221 AVX2 dispatches** on one 3-frame vector.

## NEW: an x86-only threading defect the doc did not predict

At t=8, four named vectors abort with worker `overlapping DisjointMut` panics — `00000325`,
`00000625`, `issue_48`, `…_019_vq_aom_ctest_4.2` — all four PASS on aarch64, which is clean over
three full passes at t=8/16/32. The backtrace names the site:

```
rav1d_worker_task -> rav1d_filter_sbrow_deblock_rows -> rav1d_loopfilter_sbrow_rows
  -> filter_plane_rows_y -> loopfilter_sb_direct
  -> safe_simd/loopfilter.rs:5134  (x86-only dispatcher)
  -> compact_read_per_row (picture.rs:1176)
```

**One collision has byte-identical extents** (`& _[73728..73736]` vs `&mut _[73728..73736]`), so it
is a genuine read/write race on picture pixels, not guard over-reservation — an `unchecked` build
would silently read half-written pixels instead of panicking. Seven distinct vectors have failed
across two passes and the failing set is not stable, i.e. a timing window, not content; all four
PASS in isolation over 12 targeted retries.

`#482` is ruled **out** (`RAV1D_OWNED_RECON=0` fails at a *higher* rate: 3 in 207). Still open:
who the writer is (needs `--features probe-sites`, which keeps per-record `Loc`s) and whether
removing the `check_tile` deblock barrier (A2) is the enabling change (needs the barrier put back —
the `probe-nodeblockgate` ablation went with it in `054e2ed`). **Consequence: the x86 verification
A3 demanded is not satisfied.** A3's CDEF check passes, but some other x86 site still collides and
the barrier that used to mask it is gone.

## #479 (film-grain MT) is aarch64-ONLY — measured both ways

x86 13/13 PASS at t=1/4/8; aarch64 exits 101 at t=4 and t=8 before finishing one vector.
`safe_simd/filmgrain.rs` already band-guards `(bh-1)*stride + pw` where `filmgrain_arm.rs` takes
`full_guard_mut` — so the x86 kernel has always had PR #491's shape, which independently
corroborates that narrowing is the right fix. The campaign's `--skip-group film_grain` habit was an
**aarch64-only tax**.

## Two predictions REFUTED

* **A6** ("the `CpuLevel::Scalar` hole almost certainly exists on x64") — **it does not.** All 165
  x86 dispatch gates read `rav1d_cpu_flags_mask` before summoning; `Scalar` gives 0 grants /
  337,378 refusals. So an x64 scalar-vs-SIMD A/B is valid. Bonus finding: `CpuLevel::X86V2` is
  indistinguishable from `Scalar` — nothing dispatches below the AVX2 token, so a pre-Haswell x86
  gets no vector kernels at all.
* **A1 named the wrong x86 knob.** There is no aarch64 shift constant to port (`block_shift_rule`
  is `log2(len) - 8`, arch-independent). The real x86 mis-fit is
  `#[repr(align(128))] struct Shard` + `assert!(size_of::<Shard>() == 128)`, commented "the
  M-series line size" — the guest reports `clflush size: 64`, so every shard spans **two** x86
  lines. Refit `SLOTS` / `BLOCKS_PER_SHARD`, not the shift rung.

## Confirmed

**A3** (both x86 CDEF padding loops take `offset + x_start`), **A4** (gate passes on x86 in 0.61 s
and **fails 3/3** with the in-lock `state` re-read mutated out — mutation reverted), **A5**
(`LF_BATCH_MAX = 4`, outside any arch gate), **A2**'s code removal, and **`c-ffi` on x86 Linux**
(CI's exact clippy leg exits 0 for the linux-musl target while failing on Darwin with the
documented `EAGAIN` assert — that blocker is host-OS, not architecture).

## What's in the diff

* `src/cpu.rs` + `examples/x86_tier_census.rs` + `__probe_x86tier` — the tier census. `cpu` is
  `pub` only under that feature, so the public API is unchanged.
* `scripts/x64/build_and_run_x86.sh` — the reproducible recipe (zig-cc cross-build on the host,
  run in the colima x86_64 VM), with the two link gotchas documented.
* `docs/X64_APPLICABILITY.md` section F, `benchmarks/x64_verify_2026-08-10.*`.
* No behaviour change: `cargo clippy --release --all-targets` emits exactly the same 95
  warning/error lines as pristine `main`, none from the new files.


lilith:
Filed #494 for the x86_64 t=8 deblock/compact-read race found here (with the backtrace, the seven failing vectors, and the two attribution experiments that are still open). Related: #449, same family at the CDEF tap window.

lilith:
**Closed in favour of #507, which lands this record on `main` with the one part §G superseded corrected in place.**

Verified on the remote, not asserted:

```
$ git merge-base --is-ancestor 7b223cf origin/main && echo VERIFIED
VERIFIED: 507 on origin/main
```

Why it went through a new branch rather than a rebase of this one: `verify/x64-applicability` is
checked out in another agent's worktree (`~/work/zen/rav1d-safe--x64verify`), and AGENT_BRIEF §1
says never touch another agent's branch. Nothing here was force-pushed.

## What landed unchanged

Section F in full — F0's TCG scope limits, F1's tier census (384,221 AVX2 dispatches; `Scalar` 0
grants / 337,378 refusals), F2's corpus tables, F3's film-grain both-ways measurement, F4's per-row
A1-A6 verdicts, F5's zero-MISMATCH verdict, F6's call-site census, F7's not-verified list, F8, and
F9's citation of the i265 record — plus all four `benchmarks/x64_verify_2026-08-10.*` files,
`scripts/x64/build_and_run_x86.sh`, and the `__probe_x86tier` census. #497's own record says
reproducing its x86 arms needs that script and that it deliberately did not duplicate it, so it was
the missing piece.

`main` had also been referencing §F since #497: `docs/X64_APPLICABILITY.md:189` reads *"Sections F
and G were written by different sessions. **F** (PR #493) is the first x86_64 execution pass"*, and
§G/§H are titled "the t=8 race **F found**" / "the two x86 misfits **F flagged**" — against a file
with no §F in it. That is now fixed.

## What was corrected, dated, not deleted

* **F2's "What is NOT established" list** — all three bullets are answered by §G (#497, `b700489`)
  and **the defect is FIXED**. Both writers named (`owned_recon.rs:937:42` `stitch_sbrow` and a
  second LF compact read at `safe_simd/loopfilter.rs:5134:44`); the `054e2ed` barrier removal *is*
  the exposure, and restoring it is rejected as the remedy (2.19x at t=8 to paper over a 3-row
  over-read); and **aarch64 was not merely lucky** — its dispatcher returns `false` and lets
  `LfBlock::open` size each rectangle from the group's own `wd`, so only x86 ever used the
  plane-worst-case constant. The failed `probe-sites` reproduction recorded here is kept, because it
  is what sent §G to the right instrument (`-C debug-assertions=on`).
* **F2's A2/A3 consequence** — narrowed: the coarse site was found and fixed; the general warning
  stands and §G restates it as a `debug_assert!`.
* **F4's A2 row** — same update inline.
* The `.meta` gets a header note so its four ERROR rows are not read as a live defect.

## What was NOT corrected, because it is still true

**F7's CI gap is open.** `ci.yml`'s corpus leg still runs `decode_md5_verify` at `threads = 1`
(line 141) and is still `continue-on-error: true` (line 142). The x86 `--threads 8` leg F7 asks for
does not exist, and it remains the cheapest thing that would have caught F2. Left as an open item
rather than smuggled into a records landing.

**Nothing was re-measured for this landing** — another agent holds the box for timed work — so
every number above is #493's or #497's own, cited.

Issue #496, opened 2026-08-10
x86: CpuLevel::X86V2 is inert in a default build — no safe-SIMD tier below AVX2


lilith:
## Summary

`CpuLevel::X86V2` sets `SSE2 | SSSE3 | SSE41` in `rav1d_cpu_flags_mask`, but **nothing under
`src/safe_simd/` reads those three flags.** Every x86 dispatcher gates on
`crate::src::cpu::summon_avx2()` (or `summon_avx512()` / `summon_avx512x()`), so in a default
build the safe-SIMD tier ladder starts at **AVX2** and:

* `CpuLevel::X86V2` is behaviourally identical to `CpuLevel::Scalar`;
* a pre-Haswell x86-64 CPU (no AVX2) gets **no vector kernels at all** — it runs the scalar
  reference for the loop filter, CDEF, itx, mc, film grain and loop restoration.

Found while verifying `docs/X64_APPLICABILITY.md` on real x86 (PR #493 measured the tier census:
`Scalar` gives 0 grants / 337,378 refusals, and `X86V2` is indistinguishable from it).

## Where it is not inert

* `--features asm` links dav1d's SSSE3/SSE4.1 assembly, which the mask does gate.
* `--features unchecked` uses SSE2 intrinsics in msac (`src/managed.rs::is_unchecked`).

So the level is meaningful for those builds and a synonym for `Scalar` in the default one. The
doc comment now says so (`src/managed.rs`, and `docs/X64_APPLICABILITY.md` H2), which closes the
misleading-API half of this. This issue is about the coverage half.

## Why it might matter

x86-64-v2 is the floor for a lot of deployed hardware and for distro baselines. The aarch64 side
has no equivalent hole — NEON is mandatory on aarch64, so `CpuLevel::Neon` is the real baseline
and it is fully served. On x86 the equivalent baseline is served by the scalar reference.

## What would close it

An SSE4.1-tier (128-bit) dispatch rung for the kernels where it is worth it, gated on
`simd_enabled(CpuFlags::SSE41)` before summoning a 128-bit token, in the same shape as
`summon_avx2()`. The `magetypes`-style generic-width kernels would give most of this for the
kernels already written against generic lane types; the hand-written AVX2 kernels would each need
a 128-bit sibling.

Not obviously worth doing on RD grounds — that is the decision this issue is asking for, and it
should be made against a measurement on a real pre-Haswell (or `-C target-feature=-avx2`-masked)
[...]

PR #497, opened 2026-08-10
fix(#494): the x86 loop-filter V window read 3 rows past its superblock row — sized from the plane, not the mask

### LINKED ISSUE #494: x86_64 t=8: deblock compact_read_per_row races a concurrent write (identical extents) — 4/768 vectors abort, aarch64 clean
## Summary

On **x86_64** at `--threads 8`, `main` @ `5e9975f` aborts 4 of 768 dav1d-test-data vectors with a
worker `overlapping DisjointMut` panic (surfaced as `decode error: generic error`). aarch64 passes
all four, and is clean over three full corpus passes at t=8, t=16 and t=32 on the same commit.

**One of the collisions has byte-identical extents, so this is not guard over-reservation — it is a
genuine read/write race on picture pixels.** In a `--features unchecked` build there is no panic:
the read simply returns half-written pixels.

## Failing set (set-diffed BY NAME against the aarch64 t=8 run)

| vector | x86_64 t=8 | aarch64 t=8 |
|---|---|---|
| `8-bit/data/00000325` | ERROR | PASS |
| `8-bit/data/00000625` | ERROR | PASS |
| `8-bit/issues/issue_48` | ERROR | PASS |
| `8-bit/vq_suite/Syntax_AV1_mainb8ss420_432x240_019_vq_aom_ctest_4.2` | ERROR | PASS |

A second pass with `RAV1D_OWNED_RECON=0` failed on `00000557`, `00000580`, `00000624` — **3 in its
first 207 vectors**, a higher rate. Seven distinct vectors so far and the failing set is not
stable, i.e. a timing window rather than content. All four of the first set PASS at t=8 in
isolation over 12 targeted retries, so the window needs the sustained many-decoder run.

## Site

```
rav1d_worker_task                 src/thread_task.rs:1370
rav1d_filter_sbrow_deblock_rows   src/recon.rs:3753
rav1d_loopfilter_sbrow_rows       src/lf_apply.rs:711
filter_plane_rows_y               src/lf_apply.rs:439
<loopfilter Fn>::call             src/loopfilter.rs:234
loopfilter_sb_direct              src/loopfilter.rs:103
loopfilter_sb_dispatch            src/safe_simd/loopfilter.rs:5134   <-- x86-only file
compact_read_per_row              include/dav1d/picture.rs:1176
slice                             include/dav1d/picture.rs:729
```

The deblock task's compact read, from the x86 AVX2 dispatcher's `use_compact` arm
(`tile_threading_active()`), which takes one guard per row of `cw` pixels — `cw = max_iter * 4` for
a V run (up to 128) and `tap_before + tap_after` for an H run (8-16). Both widths appear in the
panics. `src/loopfilter.rs:101` routes x86 here and aarch64 to
`loopfilter_arm::loopfilter_sb_dispatch`, so the colliding read is issued by arch-specific code
even though `compact_read_per_row` is shared.

## Extents

Default arm:

```
current:    & _[82432..82560] (128 B)   existing: &mut _[82488..82496]  (8 B)
current: &mut _[73844..73848]   (4 B)   existing:    & _[73728..73848] (120 B)
current:    & _[16640..16752] (112 B)   existing: &mut _[16672..16688] (16 B)
current:    & _[49536..49660] (124 B)   existing: &mut _[49536..49568] (32 B)
```

`RAV1D_OWNED_RECON=0` arm, with `RUST_BACKTRACE=1`:

```
current:    & _[73728..73736]   (8 B)   existing: &mut _[73728..73736]  (8 B)   <-- IDENTICAL
current:    & _[73728..73744]  (16 B)   existing: &mut _[73728..73732]  (4 B)
```

## What is ruled out, and what is not

* **#482 (tile-owned recon) is ruled OUT** — `RAV1D_OWNED_RECON=0` does not suppress it and fails
  at a higher rate.
* **Not emulation noise** — the extents and the backtrace are self-consistent and the same site
  recurs across independent passes.
* **NOT yet known: who the writer is.** The `existing:` record prints no location in a release
  build. A `--features probe-sites` build keeps per-record `Loc`s and will name both sides; one
  358-vector pass with that build did not reproduce (the rate is ~1%/vector, so expect to need
  several).
* **NOT yet known: whether removing the `check_tile` deblock barrier is the enabling change.** The
  removed barrier gated *reconstruction* of sbrow N on deblock progress reaching N-1 — exactly the
  ordering whose absence would let a tile's recon write land in rows a deblock task is reading —
  and `docs/X64_APPLICABILITY.md` A3 already records that the barrier had been masking this class.
  There is no ablation switch left (the `probe-nodeblockgate` family went with the barrier in
  `054e2ed`), so attribution needs the barrier put back temporarily.
* **NOT yet known: whether aarch64 is merely lucky.** `compact_read_per_row` is shared;
  `--features __probe_bounds` (`docs/BOUNDS_MAP.md`) prints each site's distance to the nearest
  concurrently-live foreign WRITE and is the designed instrument for deciding that.

Related: #449 (the same family at the CDEF tap window, aarch64/Windows-ARM).

## Reproduction

```
scripts/x64/build_and_run_x86.sh build --example md5_inventory
scripts/x64/build_and_run_x86.sh run ./target/x86_64-unknown-linux-musl/release/examples/md5_inventory --threads 8
```

Executed in a colima QEMU-TCG x86_64 Linux VM (4 vCPU) — correctness only, no timing. Record:
`benchmarks/x64_verify_2026-08-10.{meta,_corpus.tsv.zst,_panics.txt}`, narrative in
`docs/X64_APPLICABILITY.md` section F2. PR #493.

**A single `ubuntu-latest` CI leg running `md5_inventory --threads 8` would have caught this.**
There is no x86_64 corpus job today.

lilith:
Closes #494. Base for every measured arm: `main` @ **14e5cf4** (the branch is since rebased onto
`97755cc`, which added `perf/ctx-tl-*`; those touch `ctx.rs`, not the filter).

## What is INCOMPLETE or NOT COVERED — read this first

1. **wasm32 has the same defect in a worse form and is NOT fixed.** Its dispatcher takes ONE
   **mutable** guard reaching **16 rows** below a V edge. There `reach_after` is both the window and
   the scalar-fallback predicate, so narrowing it without the `win_*`/`reach_*` split the x86 arms
   now have could change wasm output bytes — and nothing on this box can execute wasm32 threads to
   check. Documented at the site with the exact recipe.
2. **AVX-512 is unexecuted.** The guest has no AVX-512, so the `x16` ZMM V-kernel path never ran.
   It consumes the same `(buf, base, strideb)` window as the x8/x4 paths, so a mis-sized rectangle
   would panic rather than corrupt — but it is not exercised here.
3. **No timing number, and none is derivable.** Every x86 arm ran under QEMU-TCG: no cache model,
   no store buffer, no PMU. Whether narrowing the copy (14 rows -> 4 at level-0 edges) helps on real
   x86 is **unmeasured**.
4. **The two x86 misfits from #493 are documented, not fixed.** `Shard`'s 128-byte alignment spans
   two x86 cache lines and the steady-state path touches both (`ends[0]` is at offset 72); two pure
   layout refits are named and **neither is landed, because TCG cannot measure them**. `CpuLevel::X86V2`
   is a synonym for `Scalar` in a default build — documented, and filed as #496.
5. **Reproducing the x86 arms needs #493's `scripts/x64/build_and_run_x86.sh`**, which is unmerged.
   Deliberately not duplicated here to avoid an add/add conflict.
6. One pass of the post-fix t=8 loop was **discarded**, not counted: cargo rewrote the binary the
   running container had exec'd, and a bind-mounted mmap of a replaced file is not a trustworthy
   arm. Subsequent arms run against pinned copies cargo never writes.

## Root cause

`loopfilter_sb_direct` dispatches to a different SIMD entry point per architecture, and they sit at
different levels: aarch64's returns `false` and lets `LfBlock::open` size each rectangle from the
fused group's own `wd`; x86's handles the whole superblock edge and sizes the rectangle itself —
from a **constant**, `tap_before = tap_after = 7` for a luma V run.

The filter's real reach is `lf_reach(wd)` for the width the **mask** selected, and the mask level at
a horizontal edge is `min(log2(tx_h) above, log2(tx_h) below)` capped at 2 (`src/lf_mask.rs`,
`masks[1][by4 + y][cmp::min(ttx, btx)]`). A transform never crosses a superblock boundary, so:

| level | wd | reach | rows of headroom below the edge |
|---|---|---|---|
| 2 | 16 | 7 | >= 16 |
| 1 | 8 | 4 | >= 8 |
| 0 | 4 | 2 | >= 4 |

A mask-derived window always fits. The constant does not: **at every level-0 edge in the last 4-row
band of a superblock row it read 7 rows where 4 remain — 3 rows past the bottom of its own
superblock row**, which since `054e2ed` dropped the frame-global deblock barrier is being written
concurrently.

## The writer, named

`--features probe-sites` does not reproduce it (#493 recorded that as a measured negative). What
does, and keeps per-record `Loc`s, is a plain **`-C debug-assertions=on` release build**:
`ShardRecs::locs` and the `track_caller` propagation through `picture.rs`'s helpers are both
`debug_assertions`-gated, so it costs one non-atomic store per registration and no hash table. It
reproduced at a comparable rate (2 aborts in the first 220 vectors of `8-bit/data` at t=8, vs
release's 4 in 358) and named both sides:

```
 current:    & _[98304..98432] at src/safe_simd/loopfilter.rs:5134:44   <- the V-run compact read
existing: &mut _[98304..98688] at src/owned_recon.rs:937:42            <- stitch_sbrow, NEXT sbrow

 current: &mut _[98439..98441] at src/safe_simd/loopfilter.rs:5203:25   <- an LF diff write-back
existing:    & _[98432..98448] at src/safe_simd/loopfilter.rs:5134:44   <- another V-run read
```

Both records start at 98304 = row 256 at a 384-px stride, i.e. the first row of a superblock row —
exactly the rows a V run at the last edge of the previous one over-reads.

**Two writers, which is why #482 was correctly ruled out and yet `RAV1D_OWNED_RECON=0` made it
worse**: one pairing is `stitch_sbrow` copying the next superblock row out of the owned band; the
other is that row's own `DeblockCols` task, which does not involve owned recon at all.

## Barrier attribution: yes, and it is still the wrong remedy

`054e2ed` removed a `check_tile` predicate requiring `deblock_progress >= sby` for recon tasks, so
recon of row S waited for `DeblockRows(S-1)`. Every pairing above is downstream of it. **The barrier
was not put back** — it costs 2.19x at t=8 to paper over a 3-row over-read, and its own commit
message is the fix's specification: *"the filter length is bounded by the transform heights on both
sides, which cannot cross a superblock boundary."* True of the reference and of the aarch64 path.
False of that window.

## The fix

`src::loopfilter::lf_run_reach(is_y, mask)` returns the reach of the widest width the run's mask can
select; both BPC arms of the x86 dispatcher use it for the V window. `lf_group_wd` is extracted from
`loop_filter_sb128_rust` so the new function and its test share the driver's ladder rather than
copying it.

Two things deliberately unchanged, so output cannot move:

* the **scalar-fallback bounds predicate** still tests the plane worst case, so the SIMD-vs-scalar
  decision is bit-for-bit what it was and the narrowed window is always a subset of an extent that
  test already proved in bounds;
* the **H direction** keeps the constant: its perpendicular extent is COLUMNS of rows already inside
  this superblock row, so it has no cross-row ordering to violate, and its `tap_after` (9 luma / 5
  chroma at 8bpc) is the chunked transpose load's rounding, not a tap bound.

## The defect is now deterministic and single-threaded

`loopfilter_sb_direct` carries a `debug_assert!` — arch-independent, where the superblock geometry is
in scope — that a V run's window stays inside its superblock row. With the old constant planted, a
`--release -C debug-assertions=on` build aborts on the **second vector of `8-bit/data` at
`--threads 1`**:

```
V-run window leaves the superblock row: row 380 (+7) in a 128-row superblock row,
is_y=true, mask=[f0000000, 00000000, 00000000]
```

380 % 128 = 124 — the last 4-row band, only level 0 present. Unmutated: 358/358 PASS, 0 firings.
**A race that needed a 10-minute t=8 corpus pass to show once is now a 1.5-second t=1 abort.** That
is the reusable part.

## Gates

| gate | result |
|---|---|
| `cargo test --lib` DEBUG | 77 passed |
| `cargo test --lib` release | 77 passed |
| `run_reach_equals_the_widest_group_it_can_meet` | PASS; **teeth**: planting the old constant fails it (`left: 7, right: 4`) |
| `debug_assert!` invariant, aarch64 t=1 `8-bit/data` | 358/358 PASS, 0 firings; **teeth**: mutation aborts on vector 2 |
| `forbid(unsafe_code)` | proved ACTIVE — planted `unsafe` in `src/ablate.rs` (always compiled), error anchors on `lib.rs:13:12`; restored byte-exact (`git diff --exit-code`) |
| aarch64 corpus, base vs head, by name with the MD5 as the value | **IDENTICAL, 768 rows, at t=1 AND t=8** (766 PASS / 0 MISMATCH / 0 ERROR / 2 SKIP each); head t=1 == head t=8 |
| x86_64 corpus at t=8 | see the table below |
| `mt_stress` 1/2/4/8/16 | PASS |
| `multi_decoder_pressure.sh` (8 procs x 2 iters, t=1/2/4/8/16, 5 vectors) | PASS — every md5 matches the serial reference |
| clippy `--release --all-targets` | 95 warning/error lines, **the same count as the base source files** (measured both ways on this box) |
| `cargo fmt --check` | clean |
| `--features probe-wide` standing hazards | `wide_total = 0` on `v4k_8tile` and `v1024` at t=8; sources restored byte-exact (sha256 + `git diff --exit-code`) |
| Miri, both models, each target in isolation | see the comment below — reported per target, and a watchdog TIMEOUT is reported as a timeout, never as green |

### x86_64 (QEMU-TCG guest; correctness only)

`8-bit/data --threads 8`, one pass = 358 vectors:

| arm | ERROR | MISMATCH | overlap panics |
|---|---|---|---|
| base (pre-fix) | **4** (`00000635 00000641 00000660 00000704`) | 0 | 4 |
| head pass 1 | **0** | 0 | **0** |

Further head passes and the full-corpus base/head arms at t=1 and t=8 are running as this is
written; the final table lands in `benchmarks/x64_lf_window_2026-08-10.meta` §7 and in a comment.
The pre-fix failing set is **not stable** — #493's two passes named
`00000325 00000625 issue_48 …_019_vq_aom_ctest_4.2` and `00000557 00000580 00000624`, mine named
four others — so a single clean pass proves nothing and the count of passes is the evidence.

## Docs updated in place

`docs/X64_APPLICABILITY.md` (new §G, §H — appended with a header note so it merges with #493's §F
either way), `docs/OWNERSHIP_MODELS.md` §7d (the fourth refutation, arriving from the opposite side:
reservation == footprint and still wrong), `docs/BOUNDS_MAP.md` (the instrument prices reservations
against footprints, not footprints against NEED — the hole this fell through),
`src/thread_task.rs::check_tile`, `tracker_shard.rs`'s `align(128)` comment,
`src/managed.rs::CpuLevel::X86V2`, `CHANGELOG.md`, and
`benchmarks/x64_lf_window_2026-08-10.meta`.


lilith:
## x86_64 t=8: the by-name set-diff, with the MD5 as the value

`8-bit/data --threads 8`, base = `main` @ 14e5cf4, head = this branch, both cross-built to
`x86_64-unknown-linux-musl` and run in the colima QEMU-TCG guest. `diff` of
`(name, status, actual_md5)` — **the only four differing rows are the four repairs, and each lands
on its expected MD5:**

```
211c211
< 8-bit/data/00000635  ERROR  ERR:decode frame 24: decode error: generic error
> 8-bit/data/00000635  PASS   cb6aa637c62808b6409e8d6f1bb0b3db
213c213
< 8-bit/data/00000641  ERROR  ERR:decode frame 43: decode error: generic error
> 8-bit/data/00000641  PASS   9a9f8b9faecb563dc5f4fafc5dc0724e
228c228
< 8-bit/data/00000660  ERROR  ERR:decode frame 7: decode error: generic error
> 8-bit/data/00000660  PASS   5dae1f16721633566af1f0693a27d6ed
242c242
< 8-bit/data/00000704  ERROR  ERR:decode frame 0: decode error: generic error
> 8-bit/data/00000704  PASS   1a146612fff0c1e250d3d991493c4e42
```

354 of 358 rows identical including their MD5s. **0 repaired-and-broken**, which is the thing a
count cannot show.

## Post-fix t=8 samples so far — a race that passes once proves nothing

| arm | scope | PASS | ERROR | MISMATCH | overlap panics |
|---|---|---|---|---|---|
| head, pass 1 | `8-bit/data` | 358 | 0 | 0 | 0 |
| head, pass 2 | `8-bit/data` | 358 | 0 | 0 | 0 |
| head, full corpus | **all 768** | **766** (+2 SKIP) | 0 | 0 | 0 |
| base (pre-fix) | `8-bit/data` | 354 | **4** | 0 | 4 |

The two head `8-bit/data` arms and the full-corpus arm agree **exactly** on all 358 shared names
(`diff` clean), i.e. the two independent post-fix runs produce identical pixels.

**Every vector that has failed at any point in this campaign is PASS by name in the full-corpus
t=8 arm** — #493's first pass (`00000325`, `00000625`, `issue_48`,
`…_019_vq_aom_ctest_4.2`), its `RAV1D_OWNED_RECON=0` pass (`00000557`, `00000580`, `00000624`), my
release pass (`00000635`, `00000641`, `00000660`, `00000704`) and my debug-assertions pass
(`00000037`, `00000051`). All 18 `vq_aom_ctest_4.2` variants pass, including `_019`.

## The exposure, quantified: 0.632% of V runs, and zero after the fix

A throwaway counter beside the new assertion (uncommitted; sources restored byte-exact),
aarch64 `8-bit/data` t=1, sampled every 200,000 V runs:

```
      n runs   old (plane 7/3)   new (mask-derived)
     200,000               919                    0
     600,000             3,350                    0
   1,000,000             6,438                    0
   1,600,000            10,104                    0
```

So the over-read is **not rare in the filter — it is rare in COINCIDENCE.** Only the crossings that
land on a row a peer is writing at that instant become a panic, which is why the t=8 failure rate is
~1% per vector and the failing set is unstable across passes.

## aarch64: unchanged, proven by name

Full corpus, no `--skip-group`: base and head both 766 PASS / 0 MISMATCH / 0 ERROR / 2 SKIP at
**t=1 and t=8**, and the `(name, status, actual_md5)` set-diff base-vs-head is **IDENTICAL over all
768 rows at both thread counts**. `head t=1` vs `head t=8` also identical.

## The invariant assertion over the full corpus

`--release -C debug-assertions=on`, aarch64: **766 PASS / 2 SKIP with 0 firings at t=1 AND at t=8**
(0 overlap panics). The full corpus is what makes that cover 4:2:0 / 4:2:2 / 4:4:4 / 4:0:0 and
8/10/12-bit rather than one 8-bit 4:2:0 group. Teeth unchanged: with the old constant planted it
aborts on the second vector of `8-bit/data` at `--threads 1`.

## Also verified independently on this branch

* **A3 on x86** — both x86 CDEF padding loops (8bpc `cdef.rs:1392/1400`, 16bpc `:1851/1858`) guard
  `offset + x_start` for `x_end - x_start`, matching the scalar reference. So CDEF really is narrow
  on x86 and something else was colliding, which is what sent this at the loop filter.
* **No sibling of this defect in the x86 tree** — every other x86/portable `safe_simd` path that
  touches picture memory goes through `block_mut` / `with_block_mut` / `narrow_guard` with the
  block's actual `(w, h)`. `loopfilter.rs` was the only one deriving a TAP-REACH window of its own.
* **`--features probe-wide`**: `wide_total = 0` on `v4k_8tile` and `v1024` at t=8 — narrowing an
  extent cannot add wide-path promotions, and it did not.
* **Full release test suite on aarch64**: 22 test binaries, **166 passed, 0 failed**, including
  `decode_md5_verify` (14), `decode_permutations` (19), `filmgrain_threads` (6),
  `tile_threading_parity`, `tile_threading_overlap`.
* **`mt_stress` 1/2/4/8/16** and **`multi_decoder_pressure.sh`** (8 concurrent decoders x 2 iters,
  thread counts 1/2/4/8/16, 5 vectors, every md5 matched to a serial reference): PASS.

## Still running / still NOT covered

Arms in flight: x86 `t=1` full corpus (the non-compact path this also narrows), x86 base full corpus
at t=8/t=1, a third and fourth head `8-bit/data` t=8 pass, a `probe-sites` registration-count
liveness pair, and the **barrier attribution arm** (pre-fix filter + `054e2ed`'s predicate restored
verbatim, to falsify "the barrier removal is what exposed this"). Results will land in
`benchmarks/x64_lf_window_2026-08-10.meta` §7.

Unchanged from the PR description: **wasm32 is not fixed** (same defect, worse form — 16 rows,
mutable), **AVX-512 is unexecuted**, **there is no timing number and none is derivable from TCG**,
and the two #493 misfits are documented rather than changed (`CpuLevel::X86V2` filed as #496).

lilith:
## Final results — CI 48/48 green, and the three things that were still open are closed

### 1. The barrier attribution was TESTED, not argued

A throwaway arm: the **pre-fix** loop filter with `054e2ed`'s predicate re-inserted verbatim
(`deblock_progress >= t.sby` for recon tasks, the extra param threaded through both `check_tile`
call sites; never committed, sources restored byte-exact).

| arm | scope | PASS | ERROR | overlap panics |
|---|---|---|---|---|
| pre-fix, **no** barrier | `8-bit/data` t=8 | 354 | **4** | 4 |
| pre-fix, **barrier restored**, pass 1 | `8-bit/data` t=8 | 358 | **0** | **0** |
| pre-fix, **barrier restored**, pass 2 | `8-bit/data` t=8 | 358 | **0** | **0** |

So the barrier removal is confirmed as the enabling change, and the test had a way to fail — a
pairing the barrier does not order would still have aborted. **It stays removed:** 2.19x at t=8
(`benchmarks/p1_barrier_2026-08-07.meta`) is not a price worth paying for a 3-row over-read the
window itself can fix.

### 2. Full-corpus x86 arms, both thread counts, by name with the MD5 as the value

| arm | scope | PASS | ERROR | MISMATCH | overlap panics |
|---|---|---|---|---|---|
| base (pre-fix) | full, t=8 | 764 | **2** | 0 | 2 |
| head | full, t=8 | **766** | 0 | 0 | 0 |
| head | full, **t=1** | **766** | 0 | 0 | 0 |
| head | `8-bit/data` t=8 x2 | 358 each | 0 | 0 | 0 |

Full-corpus `diff` on `(name, status, actual_md5)`, base vs head, **exactly two differing rows out
of 768** and both are repairs onto the expected MD5:

```
00000245  ERROR(frame  2) -> PASS e6122012caf67df41bcff9111f4b7def
00000625  ERROR(frame 17) -> PASS a6b0902bd590da52e3e12854800db9d2
```

**Five-way MD5 identity, 768 rows each:** x86 head t=1 == x86 head t=8 == aarch64 head t=1 ==
aarch64 head t=8 == aarch64 **base**. Every one of those diffs is empty. The x86 t=1 arm is not
redundant — at t=1 `tile_threading_active()` is false, so x86 takes the NON-compact single-guard
path this change also narrows, and 766/766 there is what says the narrowed guard is in bounds.

**Six pre-fix passes produced six different failing sets — fourteen distinct vectors, overlapping
only on `00000625`.** That is what a timing window looks like, and why one clean pass would have
proved nothing. All fourteen are PASS by name in the head full-corpus t=8 arm.

### 3. The narrowing is LIVE on x86, counted

`--features probe-sites`, x86_64 t=8, identical deterministic workload both arms:

| site | base | head |
|---|---|---|
| compact **read** `loopfilter.rs:5134` / `:5177` | 994,964 regs @ 20.8 B | 970,370 @ **19.3 B** |
| diff **write-back** `:5203` / `:5246` | 316,531 @ 8.3 B | 316,531 @ 8.3 B — **unchanged** |

−24,594 row guards (−2.47%) and a 7.2% smaller mean read extent, with the write site equal **to the
registration** — which is the design: the write set is the diff against `pristine`, and this change
never touched what the filter modifies. That bounds the perf direction (strictly fewer guards,
strictly fewer bytes copied) **without claiming a number, because TCG has no cache model.**

### 4. Miri, both models

**CI is the authority and both legs PASS** — Stacked Borrows 47 min, Tree Borrows 1 h 11 m, on the
`--no-fail-fast` invocation where every target reports. Local per-target corroboration on
macOS/aarch64 covers 7 of 8 targets under BOTH models; **`shard_liveness` TIMED OUT at 900 s under
both and is recorded as a timeout, never as green** (it spins on `TinyLock`, which Miri's scheduler
cannot get through in bounded steps here). `aligned_miri` / `pic_buf_overflow` have 0 tests without
their features — same as CI's own invocation; running them WITH those features is not done.

### 5. CI: 48 / 48 pass

Including `Conformance (decode permutations, x86_64)`, `Threading race gates`,
`Wide-path exclusion gate (non-vacuous)`, `SIMD Permutation Tests` on both arches, `MSRV (1.85)`,
i686 and `windows-11-arm`.

### Two LOCAL recipes fail for reasons that predate this branch

`just clippy` (no `--release`, so three tests' intentional `compile_error!` fires, plus unused-item
warnings in `tests/thread_cleanup_test.rs`, `tests/decode_md5_committed.rs`,
`examples/md5_inventory.rs` — none touched here) and `just api-doc-check` (builds the rustdoc surface
over ALL features including `asm`, whose build script cannot run on this host). CI's **Clippy** leg
passes, and `cargo clippy --release --all-targets` gives **95** warning/error lines — the same count
as the base source files, measured both ways on this box.

### Still NOT covered — unchanged

**wasm32 is not fixed** (same defect, worse form: one *mutable* guard 16 rows below a V edge; its
`reach_after` is both the window and the fallback predicate, so narrowing it without the
`win_*`/`reach_*` split could change wasm output bytes and nothing here can execute wasm32 threads).
**AVX-512 is unexecuted.** **No timing number exists or is derivable.** The two #493 misfits are
documented rather than changed — `Shard`'s two-x86-cache-line layout with both candidate refits named
and neither landed, and `CpuLevel::X86V2` filed as #496.

Full record: `benchmarks/x64_lf_window_2026-08-10.meta` plus seven `.tsv.zst` arms.

PR #498, opened 2026-08-10
bench(#455): the strided-2D record — soundness objection measured (yes, 17,704 gap writes), cost refutes it

lilith:
Settles a question this campaign accepted on theory in March 2026 and never measured.

`884b4b5` added strided 2-D overlap checking; `424cbbb` removed it one day later as "unsound strided tracking code" (*safe code can write to stride gaps the tracker declared unaccessed*); `cd9fdf5` concluded **"row slices are the only path"**. #472 re-opened it by solving the reference side independently, which would let ONE record cover an `h x w` rectangle exactly — no inter-row gaps reserved AND no per-row split.

## The answer to "does a WRITE ever land in the gap": YES

Over `8-bit/data` + `10-bit/data` at t=8 (429 vectors, 2,760 frames), with `bounds_probe::eval_rect` running at each of the 17 helpers that take per-row guards under tile threading, immediately before they take them, registering nothing:

- **34,547 hull-only collisions at 8-bit, 17,804 of them against a concurrently-live foreign WRITE** (3,923 / 2,119 at 10-bit).
- The largest pair is `picture.rs:2061:22` against **itself**, write-vs-write — two tile workers on the same picture rows at different columns, which is what AV1's column tiling produces.
- **`rect_ovl = 0` at every site.** Every hull collision is a *gap* collision; the exact rectangle never collided once. `gap_xstride = 0`, so the one genuinely inexact case in `884b4b5`'s predicate is never exercised.

So the objection splits in two:

- **Right about `884b4b5`.** It recorded the rectangle and handed out a `&mut` over the **hull** — verbatim #469's defect (§6 of `docs/OWNERSHIP_MODELS.md`) — and its predicate had a real **false negative** (a row wrapping past the stride) behind a **safe** public API. Both are now unit tests, not opinions.
- **Over-generalised to the record shape.** An exact 2-D record would permit all 34,547. Soundness is not what stops it.

## What actually stops it: the sharded tracker, and the number is `pct_row_wide`

Priced with each instance's **real** geometry (`BorrowTracker::probe_geometry` — the shipped shift is `block_shift_for(len)`, 6..14 across this corpus; an earlier draft of this probe mirrored the constant 12 and was wrong):

| site | rows_mean | row_shards_mean | saving | pct_row_wide |
|---|---|---|---|---|
| `cdef_apply.rs:121:33` | 4.03 | 2.967 | 1.36x | 0.54% |
| `owned_recon.rs:505:31` | 7.46 | 2.021 | 3.69x | 10.65% |
| `picture.rs:2061:22` | 8.42 | 3.672 | 2.29x | 27.88% |
| `loopfilter.rs:809:17` | 10.13 | 5.297 | 1.91x | **50.82%** |
| `mc_arm.rs:4741:25` | 15.62 | 6.500 | 2.40x | 70.59% |

A record must lock every shard its rows map to, and `>MAX_SHARDS_PER_BORROW = 4` promotes to the **all-shards wide path**. Saving is 1.4-3.7x registrations, not `h`; 10-bit is worse (28-91% wide). That is the mechanism behind the previously-measured 1.98x and 2.65x hull arms. Machinery on top: a shard is exactly one 128-byte line at `SLOTS = 7`, and `tracker_blockshift_2026-08-08.meta` §2 measured the cost as a core waiting for that line, not the atomics on it.

## Both directions falsified against a real decode

| arm | probe says | measured |
|---|---|---|
| control (arm off) | — | 358/358 PASS |
| `RAV1D_LF_HULL=1` | `hull_ovl = 0` | **358/358 PASS** |
| `RAV1D_RECT_HULL=1` (new, `__probe_rect_hull`) | `hull_ovl >= 17,454`, `gap_mut >= 9,013` | **196 PASS, 162 ERROR** |

The panic's extent is byte for byte the hull the probe recorded: `existing: &mut _[294912..318752]` against the logged rect `lo=294912 w=32 rows=32 stride=768`.

## Gates

- Predicate differential-tested against a brute-force byte-set oracle (378 K interval + 176 K rect-vs-rect cases). It caught **two real defects in the predicate before any decode ran**; two planted mutations are caught.
- Reconciles registration for registration with the independent `probe-sites` census — 0.011% on the corpus cell, 0.03% on 4K.
- `mutable_overlaps = 0` over 640,107 co-live reservation overlaps; all `lost_*` counters 0; `rows_capped = 0`.
- Default build untouched: 14/14 `decode_md5_verify`, clippy clean, `forbid(unsafe_code)` actively re-proved with a planted `unsafe` block.

## Verdict

**NOT BUILT.** The soundness objection is confirmed for the hull and refuted for the exact record; the cost structure refutes the exact record independently. The only two cells where the arithmetic is favourable are `cdef_apply.rs:121:33` (0.54% wide) and `owned_recon.rs:505:31` (10.65% wide) — and the second is a `Band` helper that may not need a record at all (§4d).

No wall-clock number is quoted from any build here — `__probe_bounds` publishes, fences and scans on every registration. Record: `benchmarks/strided_2d_2026-08-10.meta`.


lilith:
**Closing: every commit on this branch is already on `origin/main`, verified by SHA — not by narrative.**

```
$ git merge-base --is-ancestor 774612e origin/main   # = this branch's 87b9786
774612e IS on origin/main
$ git merge-base --is-ancestor 08a1f66 origin/main   # = this branch's 8b2f819
08a1f66 IS on origin/main
```

Both were carried in by **#500** (`perf/shard-granularity`, merged `024f997`), which rebased them
onto its own branch. A two-dot `git diff origin/main pr498` finds **nothing this branch has that
`main` lacks** — the seven residual lines in `bounds_probe.rs` are the *older* form of counters
`main` has since extended (`NCTR = 17` → more, plus `row_shards_max` / `pct_wide_c5|c8|c16`), and the
two in `picture.rs` and one in `lib.rs` are likewise superseded text.

Where each deliverable now lives on `main`:

| this PR | on `main` |
|---|---|
| `bounds_probe::eval_rect` + `mod rect_oracle` | `crates/rav1d-disjoint-mut/src/bounds_probe.rs:1326` / `:1634` |
| `DisjointMut::probe_eval_rect` | `crates/rav1d-disjoint-mut/src/lib.rs:817` |
| `benchmarks/strided_2d_2026-08-10.{meta,tsv,tsv.zst}` | all three present, byte-identical |
| the BOUNDS_MAP section | `docs/BOUNDS_MAP.md` → "The strided-rectangle counterfactual (added 2026-08-10)" |
| the OWNERSHIP_MODELS §6 revision | present |
| `__probe_rect_hull`, the `src/loopfilter.rs` / `picture.rs` probe hooks | present |

## Two things worth recording before this closes

**1. The title's `17,704` is this PR's own superseded number.** Commit `8b2f819` corrected the
8-bit `gap_mut` hand-sum to **17,804** against `strided_2d_rect_2026-08-10.tsv`, and 17,804 is what
`main` carries in all three places. Anyone quoting 17,704 from the title is quoting a figure this
branch itself withdrew. The *soundness* verdict is unchanged in every direction: `rect_ovl = 0` at
every site, so the exact rectangle never collided once — that finding stands and is the reason
`fill_hull` is confined to `!tile_threading_active()`.

**2. The title's other half — "cost refutes it" — is now superseded, and I have corrected it on
`main` rather than leaving it to read as current.** #505 (`e351b1a`) built the exact record at
`LfBlock::fill` and measured **−1.0% to −1.8% wall / up to −3.3% CPU on 5 of 6 multi-tile t=8
cells** (replicated, 10/10–11/11 signs), null on `c256x2048`; #506 (`3ed9bcd`) showed the t=1 cost
that kept it default-off is **code placement**, not the mechanism (0.9967, 7/9, against a
same-source control). The `pct_row_wide` mechanism this PR measured is kept — it is real, and it is
why the weak cells are weak — but it is necessary, not sufficient, and the successor model
`rows_mean / row_shards_mean` is refuted too (`c256x2048` has the highest ratio and is the one null).
Both `docs/BOUNDS_MAP.md` and `docs/OWNERSHIP_MODELS.md` §6 get a dated SUPERSEDED note in **#510**.

Nothing is deleted anywhere; the cost measurement stays on the record with its scope named.

PR #499, opened 2026-08-10
bench(#455): attribute the tiled t=8 scaling deficit — measure-only, do not merge

lilith:
Attribution of the tiled t=8 scaling deficit. **Measure-only** — `git diff b700489..HEAD -- src/ lib.rs include/ crates/ Cargo.toml Cargo.lock build.rs` is empty; only `scripts/perf/`, `docs/`, `benchmarks/` and one `.gitignore` line changed. **Do not merge without review**; there is nothing to ship here except the record.

Full narrative: `docs/TILED_SCALING.md`. Provenance/arms/hygiene: `benchmarks/tiled_scaling_2026-08-10.meta`.

## Why

The deficit had been measured three times and profiled **zero** times. Every profile in ~20 rounds was single-tile or t=1 — the one shape where **no tile parallelism exists** — so nothing that only appears when several tile workers and several superblock-row filter tasks are live at once could be seen. That is now a null-table row of its own: on a single-tile vector every stage is flat from t=2 to t=8 in *both* arms, so the whole effect is invisible there.

## The answer, ranked

t=8 wall gap to `dav1d --framedelay 1`, idle box (`foreign_max=2`), n=7 rounds, two-point fit, decomposed as `ideal + added_work + idle_cores`:

| term | 1024x576, 8 tiles | 4K, 8 tiles | owner |
|---|---|---|---|
| single-thread gap / t | 0.433 ms (28.4%) | 4.062 ms (46.6%) | **not a scaling problem** |
| **added work** | 0.447 ms (29.4%) | 1.831 ms (21.0%) | **100% the borrow tracker** |
| **idle cores** | 0.643 ms (42.2%) | 2.821 ms (32.4%) | ~41% tracker, rest a filter tail |

**1. All of the added-CPU half is the tracker, and removing it closes it exactly.** CPU growth t=1 → t=8: ours **+27.3% / +10.5%**; tracker compiled out **+4.4% / +3.6%**; dav1d **+3.2% / +3.1%**. Output bit-identical (8/8 md5s).

**2. One stage dominates.** Of the +2.239 ms/frame that t=4 → t=8 adds at 1024x576, `deblock_cols` is **+1.349 (60%, a 2.570x inflation of one stage)** and **+0.030** at the tracker-free ceiling. 4K: +4.365 of +7.704. Bands disjoint ([0.856..0.862] vs [2.175..2.247]).

**3. The profile names the leaf and finds a symbol that exists only at t=8.** `TinyLock::lock_slow` is 0.00% of busy self time at t=1 **and at t=2**, 1.19% at t=8; tracker bucket 5.98 → 14.58; `add_wide::<true>` 0.23 → 0.61.

**4. That lock is a bare spin, so a stage's cost is not that stage's opportunity.** Its `yield_now` is behind `__probe_lock_backoff`. Ablating deblock at t=8/1024x576 saves 0.250 ms of CPU and ablating CDEF 1.811, but ablating **both** saves 4.978 — super-additive; with the tracker out the same arms are additive to 0.6%. Symbol-level: removing the deblock work drives `lock_slow` **1.45% → 7.03%** and makes `add_contended` appear from zero, while every loop-filter symbol goes to exactly 0.00.

**5. A second, independent deficit survives at the ceiling:** the post-tile filter tail — 34.1% of wall at a mean **3.22 of 8** workers (17.5% at 2.59 untracked), ~9–12% of wall structural. Instrument is `TAIL_CONC`; the axis to sweep is `sbh` vs `n_workers`.

**Not** stragglers (per-worker spread 0.14 ms on 2.30), **not** `check_tile` deferral (28 per 36 admissions, constant in thread count), **not** a uniformly starved pool (44.3% of wall at exactly 8 workers), **not** dav1d's scheduler — `src/thread_task.rs` is a port of it; what dav1d lacks is the tracker.

## Largest number found, and it is not about threading

With the tracker compiled out, our filter chain costs **2.8x (1024x576) / 4.0x (4K)** dav1d's **at t=1**, deblock alone **5.6x** (11.5 vs 2.07 ms/frame at 4K), and the ratio is flat across thread count. That is the compact copy-in/write-back the safe-guard model needs. It belongs to the single-thread campaign; flagged because this is the first round to put a number on it with dav1d on the same clock.

## Two corrections to the framing this round was given

* The prior record's **"CPU ratio 1.427 → 2.04"** mixes columns: 2.039 is the **wall** ratio; CPU goes 1.427 → 1.763. On an idle box the same pair reads wall 1.318 → 1.872, CPU 1.318 → 1.623.
* Its absolutes are load-inflated, but its **ratios held to three digits** (ours/dav1d speedup at 1024x576/t=8: 0.705 there, 0.702 here).

## Gates

* corpus **766/766 PASS + 2 SKIP at t=1 AND t=8**, no `--skip-group`; by-name set-diff with the **actual md5** as the value against `aarch64_md5_fixes_2026-08-07_final` — 0 differences in either direction, and t1-vs-t8 0 differences.
* `forbid(unsafe_code)` proven **active**, not read: a planted `unsafe { transmute }` made the build fail; restored from a `~/tmp` backup (never `git checkout --`) and verified byte-exact by sha256 over the tracked tree **and** `git diff --exit-code`.
* No Miri run and none needed: no library code changed.

## A harness bug of mine, recorded rather than hidden

The first pass used `n_hi = 24` on a 16-frame IVF. Our drivers re-decode one OBU `n` times regardless; `dav1d --limit N` stops at end of stream — so the fit divided a short total by a long frame delta and dav1d read **94.1 ms/frame instead of 152.1**, which would have put the 4K t=1 gap at 1.96x instead of 1.21x. `tiled_wallcpu.sh` now counts each stream and refuses `n_hi > n_frames` with a FATAL. Caught by disagreeing with a prior record, not by inspection.

## Next, ranked (unbuilt — this round is deliberately attribution only)

1. Make concurrent filter tasks land on **different shard lines**. The contention is between *adjacent superblock rows filtering at once*, which is exactly the geometry `block_shift_for`'s `BLOCKS_PER_SHARD = 2` (~4.3 picture rows/block, fitted on whole-frame wall) puts in one block. Re-run the shift ladder **scored on `TAIL_CONC`**, not whole-frame wall. Do **not** re-run the shard-*count* ladder (256 already measured worse).
2. Price the filter tail's structure — `OWNERSHIP_MODELS.md` §7d option 2, still unbuilt and unpriced.

## Coverage gaps

aarch64 only; 8-bit YUV420; two vectors, both all-intra key frames re-decoded (no inter, no reference management); **loop restoration and super-resolution execute zero blocks**, so a null from either here is not a result; t ∈ {1,2,4,8}; 8 tiles only; default features. The occupancy histogram counts a spinning worker as **active**, so it over-states useful work.


lilith:
**Closed. The records landed as #499 asked to be read, not as its label suggested — #512, MERGED.**

```
$ git merge-base --is-ancestor 60e5e1a origin/main && echo VERIFIED
VERIFIED: 512 on origin/main
```

`docs/TILED_SCALING.md` + all 17 record files are now on `main`.

## On "measure-only, do not merge"

That label was about the **branch**, and the branch part was already settled by the time I looked:

* its nine `scripts/perf/tiled_*` tools **were already on `main`** (carried in by #500, `54e90d8`);
* it touches **zero** code files — `git diff b700489..HEAD -- src/ lib.rs include/ crates/
  Cargo.toml build.rs` is empty, which is this doc's own opening claim and it verifies;
* and `docs/AGENT_BRIEF.md` §6 has cited **this file by name in three rows** since — the
  dav1d-scheduler row, the single-tile-proxy row, and the filter-chain row — with the file itself
  absent from `main`. A brief that sends the next session to a document that does not exist is
  exactly the failure mode the campaign's own rules name.

The one thing that genuinely should not have merged was dropped: the `.gitignore` hunk here also
**reverts** main's `test-vectors` (no trailing slash) entry, which exists so a corpus *symlink* is
ignored. Only the `__pycache__/` + `*.pyc` addition was kept.

## §§1-6 landed unchanged, because they are still the current position

The additive decomposition, `deblock_cols` at +1.349 of +2.239 ms/frame (a 2.57x inflation of one
stage), `lock_slow` at 0.00% of busy self time at t=1 **and t=2** and 1.19% at t=8, the
super-additive ablation (deblock 0.250 + CDEF 1.811, both 4.978), the 34.1%-at-3.22-workers filter
tail, and §6's filter chain at 2.8x/4.0x dav1d with the tracker compiled out — AGENT_BRIEF §6 quotes
several of these as current and nothing merged since contradicts any of them.

## §7 item 1 was corrected in place, dated — both halves of it

* **BUILT AND SHIPPED.** The shift ladder it asks for — *"re-run scored on the tail, `TAIL_CONC`, not
  whole-frame wall"* — was done: **#500** (the ladder, re-fitted on the tail), **#501** (the same rung
  swept across picture size; the axis is `rows_per_block = 2^shift / stride`, crossover between 2.1
  and 3.8 rows/block, which is why 1024x576 and 4K had disagreed), **#503** (the derived rule shipped,
  `ROWS_PER_BLOCK_MIN = 4`). Pure cost change, **zero** count change — registrations identical at
  every rung, `pct_row_wide` 72.90% → 0.00% at five sites — moving ours/dav1d **2.09-2.27x →
  1.60-1.71x** with no regression on any of 17 cells.
* **The word "contention" in its diagnosis is REFUTED** (#504). Counted directly on `c256x2048` t=8:
  contended acquisitions **0.264%** of registrations, `lock_slow` **10.7%** of the tracker's CPU, and
  a *perfect* lock worth 2.378x → 2.265x — while the registration count is **identical** at t=2/4/8
  and cost per registration **doubles per doubling of workers** (2.19 → 4.52 → 9.18 → 19.71 ns). So
  ≥89% is the **uncontended** `add`/`remove` pair paying one cross-core cache-line transfer:
  **coherence**, not waiting. Notably **this document's own §2 already pointed that way** — 1.19% of
  busy self time is not the bulk of anything — and §7 overstated it. The *geometry* insight is
  untouched and is precisely what shipped; only the "why" changes, from "they stop waiting for each
  other" to "they stop sharing a cache line".
* Also flagged there: #505's finding that cost tracks **distinct shard lines visited**, not records
  filed, so the 19.71 ns/registration figure is an average and must not be reused as a marginal price
  for a count cut.

§7 items 2, 3 and 4 stand as written — the filter tail is still unpriced with `TAIL_CONC` as its
instrument, the shard-COUNT ladder is still a do-not-repeat (#502 re-confirmed a count cut null at
1.0030 on the hard cell), and §6 is still the biggest single number in the record.

Largest file 23 KB; all 18 under the 30 KB bar. Nothing was re-measured on this box.

lilith:
**Correction to my close comment above, and to the banner it describes.** I wrote that `docs/AGENT_BRIEF.md` §6 cites `docs/TILED_SCALING.md` "in three rows". Counted: `grep -o 'docs/TILED_SCALING\.md' docs/AGENT_BRIEF.md | wc -l` = **2** — the dav1d-tiled-scheduler row and the filter-chain row. A third §6 row (the post-tile filter tail) states §4's finding but names only the `TAIL_CONC` instrument, not this file, so it should not have been counted. Three was written from recollection, not a count.

The argument is unaffected — two dangling references is the same defect as three — but the number was published in the doc that landed, so it is corrected there in place and labelled: PR #514.

PR #500, opened 2026-08-10
perf(#455): shard granularity re-fitted on the tail — and the wide path is not dead (153/frame)

lilith:
Builds the two levers the tiled t=8 attribution named and left unbuilt
(`docs/TILED_SCALING.md` §7 item 1), measures them, and corrects a committed claim.
**Do not merge yet** — §Recommendation says what is still missing.
Narrative: `docs/SHARD_GRANULARITY.md`. Provenance: `benchmarks/shard_granularity_2026-08-10.meta`.

## Incomplete / not established, first

- **`msb-5`'s WALL effect is not measured.** Its counting result is solid; its timing
  grid is VOID because I launched it before one arm's binary finished building, so four
  of seven rounds timed a missing executable. The void TSV is committed and the re-run
  command is in the `.meta`. This is the one measurement the round leaves outstanding.
- **The recommended rung is NOT flipped.** Two picture sizes on the idle box (plus
  10-bit and 2048x1152 load-tagged), 8-bit 4:2:0, all-intra, 8 tiles, one aarch64 box.
  No x86_64, no tiny/256 cells — and §2 says the error is a function of picture HEIGHT,
  so size is the axis that matters most and is the least covered.
- **The two levers turned out to be one knob** — that is a derivation, in §"What was asked
  for". A run mapping over `block >> k` *is* the block shift at `shift + k`.
- **Miri's `pic_buf_overflow` and `aligned_miri` selected ZERO tests** under default
  features. Reported as 0, not as green.
- Clippy fails on `main` too (85 errors, aarch64). Zero of them are in the files here.

## What landed

**LEVER 1 — the granularity ladder, as a rational.** `BLOCKS_PER_SHARD` becomes
`BPS = (num, den)` so a rung can be COARSER than the default, which a `usize` count
cannot express. Six rungs spanning five block shifts, liveness proven per rung by the
shifts the tracker actually used.

**LEVER 2 — `MAX_SHARDS_PER_BORROW`, and what really bounds it.** Not the tracker:
`BorrowId` must stay one word. `PAIR_BITS` is now derived as `3 + log2(N_SHARDS)` (10 at
128 shards) instead of a hardcoded 12, so five pairs cost 55 bits and fit. **This is a
live change in the DEFAULT build**, which is why the corpus gate covers the default arm.

## The correction

The tracker's doc comment says wide-path promotions are ZERO at every block shift. That
was measured on `v4k_8tile`. On the **1024x576 8-tile vector at t=8 the shipped decoder
takes the all-shards path 153 times per frame** — the cell where the filter tail is worst.
One shift coarser: zero. One shift finer: triple.

Mechanism: the rule targets a fixed BLOCK COUNT, so rows-per-block is `h/256`, while the
accesses that pay one shard line per row are a fixed 7-9 ROWS. A defect in the rule's
SHAPE, not its constant.

## Numbers (n=5, idle box, `foreign_max = 1`, bands printed)

`L1024x576_420_8b__t8`, t=8, `bps-half` vs the shipped default:

| | wall | CPU | added work | idle cores |
|---|---|---|---|---|
| plain | 3.283 ms [3.256..3.322] | 18.322 | +0.475 | +0.993 |
| **bps-half** | **2.794 ms [2.767..2.811]** | **16.556** | **+0.263** | **+0.725** |
| untracked (ceiling) | 2.344 | 14.167 | +0.068 | +0.574 |
| dav1d --framedelay 1 | 1.744 | 11.294 | +0.044 | +0.333 |

0.851x wall / 0.904x CPU, bands disjoint. Ours/dav1d **wall 1.883x -> 1.602x, CPU 1.622x
-> 1.466x**. 52.1% of the distance to the tracker-free ceiling closed, and BOTH halves of
the deficit move. At 4K: 0.983x with overlapping bands, i.e. not a claim.

**Registrations per frame are IDENTICAL at every rung** (682,489 / 5,615,688, to the
registration). The knob changes the COST of a registration, not the count — the converse
of #488.

**`pct_row_wide` goes 62.3-73.2% -> 0.00%** one shift coarser, so the strided-2D record's
refuting quantity is removed and it is now blocked on ONE priced question (the machinery)
instead of two.

## The objective itself needed correcting

Scored on the tail alone, at 4K/t=8 the `bps4` rung improves BOTH tail scalars while being
**6.2% slower on wall** — it fills the tail by moving `deblock_cols` into it. **Tail
concurrency is a valid objective only at constant work.** Score on both; where they
disagree, wall decides and the tail says why.

## Gates

766 PASS + 2 SKIP at t=1 AND t=8, no `--skip-group`, set-diff BY NAME with the ACTUAL md5
as the value — 0/0/0 against the committed baseline, on the default arm and on `bps-half`
and `msb-5`. Loop-filter window `debug_assert` green under `-C debug-assertions=on` over
`8-bit/data` at t=8. mt_stress 1/2/4/8/16, multi_decoder_pressure PASS. Miri both models,
each target in isolation. `forbid(unsafe_code)` proven ACTIVE by a planted `unsafe`,
restored byte-exact by sha256 + `git diff --exit-code`.


lilith:
Addendum pushed (`577894e`): the one measurement the PR listed as outstanding now exists,
and it **weakens the headline in a way worth reading before reviewing §1**.

### `msb-5` timed, idle box (`foreign_max = 0`), n=7, `L1024x576_420_8b__t8`

| arm | wall | band | CPU | wall/plain | CPU/plain |
|---|---|---|---|---|---|
| plain | 3.283 | [3.233..3.317] | 18.428 | 1.0000 | 1.0000 |
| **msb-5** | 3.256 | [3.228..3.289] | 18.222 | **0.9915** | **0.9888** |
| bps-half | 2.806 | [2.794..2.822] | 16.472 | 0.8545 | 0.8939 |
| msb-5 + bps-half | 2.789 | [2.761..2.811] | 16.456 | 0.8494 | 0.8930 |

The t=1 cell calibrates it: the four arms read 1.0000 / 1.0000 / 1.0038 / 1.0000 where the
true ratio is exactly 1.000 (every rung is the same code path at one thread), so the floor
is ~0.4%.

**`msb-5` alone is 0.85% of wall / 1.12% of CPU with overlapping bands** — about a point,
same sign on both, not cleanly separated from zero at n=7.

### Which decomposes the 14.6%, and corrects my own emphasis

`msb-5` removes **79%** of the all-shards promotions and buys ~1%. `bps-half` removes
100% of them **and** cuts multi-shard registrations 72,336 → 23,910, and buys 14.6%.

So **the wide path is worth roughly one point of the fourteen.** The 153 promotions per
frame in §1 are real and were denied by the tracker's own doc comment — that correction
stands — but they are not where the money is. The money is the ordinary multi-shard
registrations, i.e. how many shard cache lines a borrow touches. §1 now points at §5e so
nobody prices the wide path from its count alone. The cap adds nothing on top of the
coarser block (0.8494 vs 0.8545, inside the floor), as it must.

### Also in the addendum

- Miri `--lib` (27 tests, including the new granularity one): **27 passed under both
  Stacked and Tree Borrows**. `guard_move_release` 2/2 both models.
- `shard_liveness`'s first `rc=1` was **my invocation bug** — the target selector was
  passed as one shell word, so cargo rejected the argument and the log has no
  `test result:` line at all. Re-issued; a timeout would be `rc=124` and would be
  reported as a timeout.
- **A harness mistake worth not repeating:** the first re-run attempt queued two
  `measlock` holds against the SAME output path. `measlock` serialised the runs
  correctly, but the second truncated the file the first had written, and my reducer
  read it mid-truncation and printed `n=1` rows from two different sweeps. The committed
  TSV is the second, clean, 7-round hold.

PR #502, opened 2026-08-11
perf(#455): the LEFT neighbour context's reads leave the tracker — 10.5-13.9% of registrations, 0.9842 geomean, and the one cell it proves is not a count problem

lilith:
Campaign index: #455. Base `main` @ `414515c`. Full record with everything NOT
done first: `docs/CTX_TL_SPLIT.md` §6c; raw data
`benchmarks/ctx_read_split_*_2026-08-10.*` + `.meta`.

## What this is

The census (`benchmarks/cost_census_2026-08-10.meta`, item 4a) ranked the
remaining levers on the cells that miss 1.4x and called this one **"the cheapest
remaining count cut"** while saying in the same sentence that it would not move
screen text. Both halves reproduced.

PR #492 split `CaseSet::many`'s two neighbour directions so the worker-local
`t.l` WRITE side could use `&mut` instead of a tracker record, and `f7c4e6f`
(§6b) did the 17 direct `t.l.<field>.index(..)` reads. What was left was the
fifteen `(a: &BlockContext, l: &BlockContext)` helpers in `src/env.rs` plus
`sm_flag`/`sm_uv_flag` in `ipred_prepare.rs`, which took both directions as ONE
reference type and so dragged the worker-local one through the tracker with the
genuinely shared one. `l` now takes `&mut BlockContext` and reads through
`get_mut()`.

**Count reduction with NO extent change and no new machinery.** `get_mut` is a
field access, so there is nothing to price against the registrations removed —
the trap that sank the V batch cap (`MUT_RECON_KERNELS.md` §19e). Nothing here
widens a reservation; `BOUNDS_MAP.md` says `ctx.rs:99:27` has zero widening
headroom, and this change moves no extent by a byte.

## Read the negatives first

* **No cell crosses 1.4x against dav1d.** Geomean 0.9842 over 11 cells.
* **`c256x2048` t=8 — the ONE cell in the census whose tracker-removed ceiling
  is band-safely under 1.4x — does not move**, on either instrument (wall
  1.0030, CPU 1.0016), despite 35,283 registrations coming out. Its tracker cost
  is CONTENTION, not count: 18.78 ns/registration against 2.78-8.11 elsewhere,
  mostly `TinyLock::lock_slow` spin. **A count cut is not that cell's lever**,
  and that is the most useful thing in this PR.
* **Screen text and screen UI cannot be fixed by tracker work at all** — the
  census's tracker-removed ceilings are 2.375x and 1.584x, both above the bar.
  This PR does not pretend otherwise.
* **Three wall rows read above 1.000** (`c1024x384` 1.0045, `c256x2048` 1.0030,
  `v4k8tile` t=8 1.0052). None is significant, none has disjoint bands, and the
  CPU leg reproduces none of them — but they are printed unrounded.
* **Eleven of the sixteen converted helpers are inter-only and fire on NO timed
  cell here.** They are covered by the 766-vector md5 gate and by the
  differential test, and by no wall clock. Their population on inter content is
  unknown, not zero.
* Not measured: x86_64/wasm32/linux at run time, `unchecked`, `asm`, `c-ffi`,
  t=2/4/16, 10/12-bit, 4:4:4, inter content. One source image per screen class.

## Measured

Registrations (`--features probe-sites`, `lost = 0` on all 16 runs), per frame:
**−10.5% to −13.9% at t=1** and −5.3% to −7.4% at t=8 — the same ABSOLUTE cut at
both, because none of these sites is one of the four `tile_threading_active()`
latches. Four of the five cells the census predicted match **to the
registration**; `ui_q20` is off by 17, which is exactly the `sm_flag` early-return
asymmetry the census flagged as an assumption (measured 50.2%/49.8%).

Wall, n=9, `measlock`, idle box, round 1 discarded as cold, two-point fit, arms
rotating, paired per-round quotient + exact sign test beside the ratio of
medians:

| cell | t | base ms/f | head ms/f | ratio | bands disjoint | h<b | p |
| text_q20 | 1 | 2.656 | 2.556 | **0.9623** | yes | 9/9 | 0.004 |
| ui_q20 | 1 | 3.228 | 3.117 | **0.9656** | yes | 9/9 | 0.004 |
| text_q20 | 8 | 1.500 | 1.450 | **0.9667** | yes | 9/9 | 0.004 |
| c1024x576 | 8 | 3.256 | 3.183 | **0.9778** | yes | 9/9 | 0.004 |
| ui_q20 | 8 | 2.594 | 2.539 | **0.9786** | yes | 9/9 | 0.004 |
| c3840x256 | 8 | 5.870 | 5.787 | 0.9858 | no | 8/9 | 0.039 |
| c1024x192 | 8 | 1.456 | 1.439 | 0.9885 | no | 7/9 | 0.180 |
| v4k8tile | 1 | 300.0 | 296.9 | 0.9896 | no | 8/9 | 0.039 |
| c256x2048 | 8 | 3.711 | 3.722 | 1.0030 | no | 2/9 | 0.180 |
| c1024x384 | 8 | 2.456 | 2.467 | 1.0045 | no | 2/9 | 0.180 |
| v4k8tile | 8 | 48.14 | 48.39 | 1.0052 | no | 4/9 | 1.000 |

CPU (n=7) agrees on all five established cells and turns `c1024x192` into 7/7.
The base column reproduces the committed census to within 0.3% on seven of
eleven cells, which is the cross-check that the two harnesses agree.

### One calibration worth more than the win

`ui_q20` t=1 prices the marginal registration at **6.21 ns** here; §6b's
read-side part one measured **5.9 ns** on the same cell from a *different*
population. The marginal rate is above the census's per-cell
`(plain − untracked)/registrations` on every cell that moved, by **1.07x to
3.60x** — so that quantity is a **lower bound** on a real removal, not a point
estimate. The multiplier is NOT uniform (3.60x at `text_q20` t=1, 1.07x at
`ui_q20` and `c1024x576` t=8), so it is a per-cell fact, not a factor to carry. Where it is not exceeded
(`c3840x256`: 0.77 measured against 8.12 nominal) the cell measured null.

## Gates

| gate | result |
|---|---|
| corpus, **no `--skip-group`**, t=1 | **766 PASS + 2 SKIP**, set-diff BY NAME (key `(group, name)`, value `(status, actual md5)`) vs `aarch64_md5_fixes_2026-08-07_final.tsv.zst`: **CLEAN** |
| corpus, t=8 | **766 PASS + 2 SKIP**, set-diff **CLEAN** |
| set-diff t=1 vs t=8 | CLEAN |
| re-run after the LAST code-touching commit (`33147db`) | yes — the table above is that run |
| base-vs-head md5 on all 11 timed vectors, t=1 and t=8 | 16/16 identical |
| `cargo test --lib`, release AND debug | 80 passed / 8 ignored, both |
| loop-filter window `debug_assert`, `-C debug-assertions=on`, `8-bit/data` t=8 | 358/358 PASS |
| `mt_stress` (threads 1/2/4/8/16 x trials) | pass |
| `multi_decoder_pressure` — 12 concurrent decoders, thread counts 1/2/4/8/16 | **PASS**, every md5 equals the serial reference |
| `tile_threading_overlap` / `reproduce_overlap` / `thread_cleanup_test` | pass (9 of 11 are `#[ignore]` on `main` too — I did not touch `tests/`) |
| `#![forbid(unsafe_code)]` | proved **ACTIVE**: a planted `unsafe { transmute }` in `src/picture.rs` fails against **`lib.rs:13`**; file restored from a backup copy, sha256 `fa02c12b…` before and after, `git diff` clean |
| clippy, aarch64 and `x86_64-apple-darwin` | fails on BOTH — and **measured to be pre-existing**: the same run on the base tree fails identically, and zero findings are in `env.rs` / `decode.rs` / `recon.rs` / `ipred_prepare.rs`. The one file in head's set that is absent from base's (`examples/profile_ivf.rs`) fires on base too when clippy reaches it |

### The differential test, and its teeth

`src/env.rs::left_split_parity` runs all fifteen helpers against their pre-split
bodies **transcribed verbatim from `414515c`** over 20,000 randomised contexts,
plus the two smooth-flag helpers. Eleven were mechanical `index()` → `get_mut()`
swaps that a scripted rewrite could have transposed an index in; four were
restructured because the two directions are no longer one type.

Planted, each confirmed to FAIL, `src/env.rs` restored from a backup copy and
verified byte-exact by sha256 (`4a5e3852…`):

1. `get_comp_dir_ctx`'s left arm reading the ABOVE edge — FAILS
2. `get_tx_ctx`'s left index transposed `yb4` → `xb4` — FAILS
3. `get_filter_ctx`'s left arm on `xb4` — FAILS
4. `get_mask_comp_ctx`'s left result masked — FAILS

The oracle also caught a defect in the test HARNESS before any decode ran: the
pre-split `cnt[(ref0 - 4)]` ladders index past their arrays unless
`intra == 0` implies `ref[0] >= 0`, so the generator now encodes that coupling
and says why.

## Miri

Both aliasing models, ONE TARGET AT A TIME (a batch lets later targets never
execute and read as health). Results in the thread below when the run finishes;
`shard_liveness` is expected to time out on aarch64 and will be reported AS a
timeout.


lilith:
## CI: 22/22 pass, and the leg count is the point

Opened against `main`, so the full `ci.yml` ran — **22 checks, all pass**
(9 `Build & Test` platforms including `windows-11-arm`, `macos-15-intel` and
both i686 legs; both `Conformance (decode permutations)`; both `SIMD
Permutation Tests`; `Tile Threading`; `Threading race gates`; `Coverage`;
`Clippy`; `Format`; 3 cross-compiles). A PR opened against any other base would
have run only the path-filtered `disjoint-mut CI` and looked just as green.

`disjoint-mut CI` did NOT run, and that is correct rather than a gap: it is
path-filtered on `crates/rav1d-disjoint-mut/**`, and
`git diff 414515c..HEAD -- crates/` is **empty** — the tracker crate is
byte-identical to `main`.

## Not a public API change

`env` and `ipred_prepare` are **private** modules (`lib.rs:171`, `:176` — plain
`mod`, no `pub`), and `docs/public-api/rav1d-safe.txt` contains zero `env::`
entries. The `l: &BlockContext` -> `&mut BlockContext` signature change is
entirely internal.

## Standing hazards, replanted

`--features __probe_wide`, `crates/rav1d-disjoint-mut`'s `tests/wide_exclusion.rs`:

| plant | result |
|---|---|
| baseline | ok (0.05 s) |
| `4af62ae`'s in-lock `state` re-read deleted from `add_contended` | **FAILED** |
| `active()` cut to one shard | **FAILED** |
| after both restores | ok (0.05 s) |

Restored from a backup copy in `~/tmp`, never `git checkout --`; sha256
`4a59a6cf942337df13e5f7c3c8fef9ab3d467162bf018095cb4667e6a80c0c3b` and
`git diff --exit-code` clean, `touch`ed after each restore. **The digest is NOT
`CTX_TL_SPLIT.md` §7's `d4e03d4a…`** — `tracker_shard.rs` changed when the
shard-size round merged, so a future replant must re-derive its own baseline
rather than compare against either value.

## Miri — Stacked Borrows leg complete

One target at a time (a batch lets later targets never execute and read as
health). `cargo +nightly miri test`, aarch64 macOS.

| target | Stacked Borrows |
|---|---|
| `rav1d-disjoint-mut --lib` | 27 passed |
| `narrow_release` | 1 passed |
| `soundness` | 25 passed |
| `wide_exclusion` | 1 passed |
| `guard_move_release` | 2 passed |
| `pic_buf_overflow` | **0 tests ran** — feature-gated, selects nothing here. Reported as 0, never as green |
| `aligned_miri` | **0 tests ran** — same |
| `shard_liveness` | **TIMEOUT (900 s, rc=124)** — the target the brief warns about on aarch64. Reported AS a timeout. CI's Linux legs cover it |
| **`rav1d-safe --lib env::left_split`** | **3 passed** (521 s) |

The last row is the one that is new: it puts the actual `&mut BlockContext` +
`get_mut()` path this PR introduces under an aliasing model rather than merely
compiling it. It runs 200 trials under `cfg(miri)` instead of 20,000, and
`trial_floor_is_not_vacuous` asserts in a NORMAL build that 200 still binds
every liveness assertion the big test makes — so the Miri leg cannot quietly
weaken on its own.

Tree Borrows is running; results to follow.

lilith:
## Miri — both models complete

| target | Stacked Borrows | Tree Borrows |
|---|---|---|
| `rav1d-disjoint-mut --lib` | 27 passed | 27 passed |
| `narrow_release` | 1 passed | 1 passed |
| `soundness` | 25 passed | 25 passed |
| `wide_exclusion` | 1 passed | 1 passed |
| `guard_move_release` | 2 passed | 2 passed |
| `pic_buf_overflow` | **0 tests ran** | **0 tests ran** |
| `aligned_miri` | **0 tests ran** | **0 tests ran** |
| `shard_liveness` | **TIMEOUT (900 s, rc=124)** | **TIMEOUT (900 s, rc=124)** |
| **`rav1d-safe --lib env::left_split`** | **3 passed** (521.05 s) | **3 passed** (520.93 s) |

The eight tracker-crate rows are the CONTROL — `git diff 414515c..HEAD -- crates/`
is empty, so they can only reproduce `main`. The ninth is the new one, and it is
what makes this a real Miri leg rather than a formality: it runs this round's
`&mut BlockContext` + `get_mut()` path under both aliasing models instead of
merely compiling it.

Reported as measured, not rounded up:

* `pic_buf_overflow` and `aligned_miri` select **zero** tests under these
  features. **0, never "green."**
* `shard_liveness` **timed out in both models** — `rc=124` at 900 s, the target
  the campaign brief warns about on aarch64. CI's Linux legs cover it. (An
  `rc=1` with no `test result:` line would be an invocation error and a
  different thing.)
* The 200-trial `cfg(miri)` floor is not a coverage relaxation, and the Miri
  pass is the proof: `all_fifteen_helpers_…`'s liveness assertions are
  UNCONDITIONAL, so a 200-trial run that failed to reach all four
  `have_top`/`have_left` combinations or all five `get_comp_dir_ctx` outputs
  would have FAILED here rather than passed with less coverage.

## CI final

**22/22 pass** on `7640a14` (and again on `18e0dc4`, docs-only). Nine
`Build & Test` platforms including `windows-11-arm`, `macos-15-intel` and both
i686 legs; both `Conformance (decode permutations)`; both `SIMD Permutation
Tests`; `Tile Threading`; `Threading race gates`; `Coverage`; `Clippy`;
`Format`; three cross-compiles.

## One correction I made to my own numbers

The first version of this PR said the marginal registration is worth "1.6-3.6x"
the census's per-cell rate. That is wrong for two cells I printed in the same
table (`ui_q20` t=8 is 5.05 vs 4.72 = **1.07x**). Recomputed across all seven
cells the true range is **1.07x to 3.60x**, and the corrected statement is the
more useful one: the multiplier is largest exactly where the census rate is
smallest, so it is a per-cell fact and not a factor to carry. Fixed in the PR
body, `CTX_TL_SPLIT.md` §6c-7 and `AGENT_BRIEF.md` §2 (commit `36c2c0b`).

PR #503, opened 2026-08-11
perf(#455): ship the derived rows-per-block shard rule — 2.09-2.27x -> 1.60-1.71x of dav1d on the cells the campaign was mis-quoting

lilith:
**The derived rows-per-block shard rule is now the DEFAULT.** The block-count
rule it replaces is `--features bps-blocks`, the base arm.

`docs/BPS_ROWS_DEFAULT.md` · `benchmarks/bps_rows_default_2026-08-11.{meta,txt}`

## Why now

#501 built this rule, measured it as the only arm that regresses no cell across
17 picture sizes, and left it **default-off** for one stated reason: only two
sizes had been tested and the error is a function of picture HEIGHT. Its own
17-cell height sweep was that coverage.

Leaving it off cost more than time. `SHARD_SIZE_SWEEP.md` §5d tabulates
**1.61x / 1.64x / 1.70x** of dav1d on the small-height tiled cells — the GATED
ARM's figures. Shipped `main` measured **2.06x / 2.24x / 2.29x** on the same
cells six hours later (`cost_census_2026-08-10.meta`). Both records are honest;
the arrangement was the foot-gun. After this they are the same build.

## Measured — the SHIPPED build, twice

n=7 (8 rounds, round 0 dropped as cold), paired per-round ratios, `measlock`,
idle box, dav1d 1.5.4 `--framedelay 1` interleaved in the same sweep, 8 tiles,
t=8. Re-measured end to end on binaries rebuilt at the last code-touching commit;
the first pass agrees within 1.1% on every cell and is committed beside it.

| cell | head/base | [min..max] | base/dav1d | **HEAD/dav1d** |
|---|---|---|---|---|
| c1024x192 | **0.7656** | [0.7594..0.7698] | 2.093 | **1.601** |
| c1024x384 | **0.7530** | [0.7374..0.7678] | 2.268 | **1.708** |
| c3840x256 | **0.7473** | [0.7287..0.7886] | 2.195 | **1.649** |
| c1024x576 | **0.8889** | [0.8637..0.9022] | 1.825 | **1.616** |
| c1024x576 **10-bit** | **0.9000** | [0.8907..0.9144] | 1.834 | **1.645** |
| c512x576 | 0.9732 | [0.9071..0.9916] | 1.812 | 1.734 |
| c256x2048 | 0.9921 | [0.9552..1.0175] | 2.358 | 2.323 |
| v4k8tile | 1.0000 | [0.9937..1.0093] | 1.389 | 1.395 |

All seven paired rounds below 1.0 on the first six. CPU 0.87 / 0.85 / 0.87 /
0.93 / 0.93 on the same cells.

**No cell regresses.** The last two are **identity controls**: the rule provably
computes the SAME shift there on every plane (read off the tracker, not
predicted), so head and base are the same decoder and their spread IS this
grid's noise floor — ±1.0% and ±3.1%. Every cell at t=1 is the same kind of
control (the rule is gated on tiles ≥ 2 and threads > 1) and reads 0.9972-1.0004.

## The 512x576 anomaly does not reproduce

#501 read **0.995** there for the derived rule against 0.927/0.930 for the two
rungs it sits between — non-monotone in coarsening, and its stated blocker. The
rule and the ladder both move luma and chroma together, so no arm can say which
plane is responsible; `--features probe-shiftpin` pins each plane independently.
A 3x3 factorial, n=7, idle box, paired vs the block-count rule:

| luma \ chroma | 8 | 9 | 10 |
|---|---|---|---|
| 10 | 1.0000 (base) | 0.9625 | 0.9918 |
| 11 | 0.9418 | **0.9213** `bps1` | **0.9282** the derived rule |
| 12 | 0.9410 | 0.9184 | **0.9336** `bps-half` |

The derived rule sits **between** the two rungs, exactly where monotonicity says.
Additivity residuals within ±2% on wall and CPU: **the planes do not interact**,
which was the only mechanism on offer. The unpinned default build in the same
sweep reads 0.9244, cross-checking its pinned twin to 0.4%.

The real residual, now isolated: **one rows target serves both planes** while a
4:2:0 chroma tap window is half the luma one. Chroma +1 is worth 3.75% and
chroma +2 gives most of it back, at all three luma shifts — **~0.7%** at this
cell. Named as a follow-up, with the instruments to price it.

## Counted, no clock

* **Registration counts identical between arms on every cell.** The knob changes
  the COST: 5.8-9.3 ns/registration → **3.2-3.9**, which is the cost census's
  *uncontended* rate. The 4.5-6.4 ns band is again refuted as a cross-cell
  constant (3.64-19.40 here).
* **`w_full` = 0 on all 32 rows** — the coarser block does not trade
  wide-by-shard-count for wide-by-slot-exhaustion, including where it coarsens
  by three shifts. `SLOTS` untouched.
* **Corrects `SHARD_GRANULARITY.md` §1**: the 153 all-shards promotions/frame at
  1024x576 are **not on a picture plane**. Two shifts on both planes leaves them
  at 3,060/20f; `--features bps-1` (one shift on *every* instance) takes them
  to 0. Coarsening is monotone, so the promoting buffer has no stride to declare.

## Gates

Corpus **766/766 at t=1 AND t=8 on the DEFAULT arm**, no `--skip-group`,
set-diffed BY NAME with the actual md5 as the value — CLEAN, plus a `bps-blocks`
leg and a t1-vs-t8 leg. Debug-assertions `8-bit/data` t=8: 358/358.
`mt_stress` 1/2/4/8/16 × 5. `multi_decoder_pressure` 12 concurrent decoders.
13 tracker feature configurations. `cargo test --lib` release and debug.
Every CI clippy/doc/fmt leg rc=0; `--all-targets` fails identically on the base
commit (81 aarch64 errors each) with zero findings in any file this branch
touches.

Teeth, planted and restored from backup copies (never `git checkout --`),
verified byte-exact by sha256 + `git diff --exit-code`:

| plant | rule test | seam test |
|---|---|---|
| `ROWS_PER_BLOCK_MIN` 4 → 1 | **FAILS** | **FAILS** |
| `set_row_stride` back to a no-op | passes | **FAILS** |
| `ROWS_RULE_ACTIVE` forced false | **FAILS** | passes (reported as a gap) |

`forbid(unsafe_code)` proven ACTIVE at **`lib.rs:13:12`**; the two standing
`wide_exclusion` hazards replanted and both FAIL.

**Two gate-hygiene corrections** worth more than this PR: 9 of the 11
overlap/cleanup tests are `#[ignore]`d, so the invocation prior rounds recorded
as passing ran **2 of 11** (all 11 pass with `-- --ignored`); and a fresh
worktree has neither `test-vectors/dav1d-test-data` nor `bench/photo_4k.avif`,
which made the first corpus pass here exit `rc=101, lines=0`.

## What is NOT done

* **No x86_64 leg** (clippy target only). Sign should carry, magnitude should not.
* **No held-out size for `ROWS_PER_BLOCK_MIN`** — still fitted on #501's grid.
  `--features rpb-2 / rpb-8 / rpb-16` is the new ladder for re-fitting the
  constant that actually ships, and it was **not swept here**.
* No 4:4:4, 12-bit, film grain, or inter content. One box, one content class.
  Loop restoration executes zero blocks on every cell.
* `msb-5` not bundled (~1%, not separated from zero).
* `c256x2048` is the worst cell in the grid at 2.32x and is now refused by two
  levers — #502's count cut and this granularity cut. 19.16 ns/registration
  against 3.2-3.9 elsewhere. It needs a third.
* Miri locally: 7 of 8 targets clean in all four (model x rule) configurations;
  **`shard_liveness` TIMES OUT (rc=124, 900 s) in all four** and is reported as
  a timeout, never as green — under Tree Borrows it does not finish even its
  first test. `docs/BPS_ROWS_DEFAULT.md` §8c, and the comment below. CI's Linux
  Miri legs are what cover that target and they are green.


lilith:
## Miri — both aliasing models x both rules, each target in isolation

7 of 8 targets clean in all four (model, rule) configurations. Full table in `docs/BPS_ROWS_DEFAULT.md` §8c; raw in `benchmarks/bps_rows_default_miri_2026-08-11.tsv`.

| target | SB default | SB `__bps_blocks` | TB default | TB `__bps_blocks` |
|---|---|---|---|---|
| `--lib` | **29** | 28 | **29** | 28 |
| `narrow_release` | 1 | 1 | 1 | 1 |
| `soundness` | 25 | 25 | 25 | 25 |
| `wide_exclusion` | 1 | 1 | 1 | 1 |
| `guard_move_release` | 2 | 2 | 2 | 2 |
| `pic_buf_overflow` | **0 tests ran** | 0 | 0 | 0 |
| `aligned_miri` | **0 tests ran** | 0 | 0 | 0 |
| `shard_liveness` | **TIMEOUT 900 s** | **TIMEOUT** | **TIMEOUT** | **TIMEOUT** |

29 vs 28 on `--lib` is the liveness proof, not a discrepancy: the extra test is `rows_rule_targets_picture_rows_not_block_count`, which a compiled-in rung deliberately `cfg`s out. The shipped rule is *exercised* under both models, not merely compiled.

**Reported as timeouts, never as green.** `rc=124` at 900 s in all four. Under Stacked Borrows two of `shard_liveness`'s five tests complete first (`concurrent_disjoint_is_never_refused`, `concurrent_overlaps_are_caught`, under both rules) and the run stops inside `exhaustive_pairs_match_the_interval_predicate`; **under Tree Borrows 900 s does not finish even the first test**, so TB coverage of that one target is zero, not partial. CI's Linux Miri legs are what cover it — and they are green on this PR.

lilith:
**CI green on `6b57d57`: 35/35, BOTH workflows.** `CI` success + `disjoint-mut CI` success — the tracker change is path-filtered into the second one, so a run with only ~14 checks would have tested a fraction. Counted: 35 on the deduplicated head run, 48 while the push- and PR-triggered runs overlapped.

The Linux Miri legs are the ones that matter here, because `shard_liveness` **times out locally on aarch64** in all four (model x rule) configurations. On CI they run to completion: Stacked Borrows 50 min, Tree Borrows 71 min, both success — first on `9f4027d` (`disjoint-mut CI` push run 31472146663) and again on `6b57d57`. Everything after `c2b5d0f` is docs/benchmarks/scripts only, so those two Miri passes are on byte-identical decoder code (`git diff c2b5d0f HEAD -- src crates include lib.rs Cargo.toml build.rs` is empty).

Not merging.

PR #505, opened 2026-08-11
perf(tracker): exact strided-rectangle borrow records — measured, default-off, and the cost model they correct

lilith:
## What this is

The **exact strided-rectangle borrow record** — one registration describing `h`
row segments with **no inter-row gap reserved**. It is the third shape after the
per-row split and the refuted hull, and it is the mechanism `docs/AGENT_BRIEF.md`
§6 and `docs/C256_CONTENTION.md` §9 named as the remaining direction for
`c256x2048` at t=8.

It works, it is exact, it is sound at t=8, it is live, and **it is not the
default**. Full record: `docs/RECT_RECORDS.md`,
`benchmarks/rect_records_2026-08-11.{meta,*.tsv}`.

## The result, worst thing first

| where | wall | CPU | sign | replicated |
|---|---|---|---|---|
| **`v4k8tile` t=1 — where the path NEVER executes** | **+0.9% to +1.3%** | same | **0/11, twice** | yes, grids D+E |
| **`c256x2048` t=8 — the cell it was built for** | −0.2% to −0.5% | −0.6% | **4/7, 5/10, 8/15 — coin flip** | yes, three grids |
| multi-tile t=8, 5 of 6 other cells | **−1.0% to −1.8%** | −1.3% to **−3.3%** | 6/7 to **11/11** | yes, grids B3+E |

Both controls — a **byte-identical copy** of the base binary and a
**same-source-different-build layout control** — sit within ±0.5% of 1.000 with
coin-flip signs on every one of those cells, so the t=8 win and the t=1 loss are
both above the floor.

**Why it is default-off:** it costs ~1% at t=1 on 4K, where this decoder's
largest gap already lives (the filter chain alone is 4.0x dav1d there), to buy
−1.5% at t=8. At t=1 the rectangle path cannot execute, so that cost is code
size — and moving the attempt out of line (`#[inline(never)]`, the obvious fix
for a function that is `#[inline(always)]` and monomorphised 12 ways) moved it
1.0103 → 1.0088 and left the sign at 0/11. **The t=1 cost is measured and NOT
explained.** Next step is `cargo asm` / `llvm-lines` on `fill`, not more timing.

## The number the round exists for

**A `LfBlock::fill` per-row registration costs 2.42-2.71 ns at the margin.**
Measured by ADDING a duplicate per row in ONE binary (`RAV1D_LF_DOUBLE`,
+180,434 registrations/frame, +3.4% wall, **0 of 25 rounds on the other side**).
The cell's AVERAGE is **19.71 ns/registration**.

So the largest registration site in the decoder — **31.7% of the population** —
is **3.9-4.4% of the tracker's CPU**, and removing 89% of it could never have
been worth more than ~1.5% of wall. **`docs/AGENT_BRIEF.md` §6's "one cross-core
transfer per registration" is wrong as a marginal price**: `fill` files 8.98
records per **2.09** distinct shard lines, and a repeat touch of a line this core
already owns is ~7.7x cheaper than the average registration. **The cost tracks
DISTINCT SHARD LINES visited, not records filed — so a count cut at the DENSEST
site is the least valuable count cut available**, which is exactly why the
densest site produced the smallest win. That closes the "registration density"
family as the fifth refused lever on this cell.

## How it is exact, and why it is neither refuted shape

* **Storage is free.** The record keeps the rectangle's **hull** in the two words
  a plain interval already used, and `(rows, seg)` is *recovered* from the hull
  and the instance's declared row stride: `span = (rows-1)*s + seg` with
  `1 <= seg <= s` is a bijection. `Shard` stays **exactly 128 bytes**, no side
  table, no second cache line. `mutable: u8` became `flags: u16` (high byte =
  rectangle) so `alloc`'s empty-shard arm still publishes both bitmaps with ONE
  store.
* **Not a hull.** Overlap detection walks rows and knows nothing of the gaps, so
  a concurrent writer in a gap — two tile columns of the same picture rows, the
  routine case — is not reported. That is the false positive that confines
  `fill_hull` to `!tile_threading_active()`.
* **Not the reverted March-2026 strided tracker.** That one had an exact record
  and a reference over the whole hull: UB under both aliasing models.
  `DisjointImmutRectGuard` has **no `Deref`**; `row(r)` derives that row from the
  buffer pointer, so no reference wider than one row ever exists.
* **Nothing is approximated.** `add_rect` **declines** — and the caller runs its
  per-row loop unchanged — with no declared stride, a stride mismatch,
  `seg > stride`, `rows > MAX_RECT_ROWS`, a hull spanning more than
  `MAX_SHARDS_PER_BORROW` blocks, a full shard, or a live wide record. Declining
  rather than promoting is why the wide list needs no rectangle support: a
  promoted rectangle would degrade to its hull and could then refuse a legitimate
  borrow.
* **The hot `find` loop is untouched.** Its hit is a PREFILTER; the caller passes
  it through `refine`, one load and one branch when the shard holds no rectangle
  record, inside a branch that is essentially never taken.

## Gates

* **corpus 766 PASS + 2 SKIP, mismatch=0 error=0, in BOTH arms × t=1 and t=8**,
  no `--skip-group`; set-diff **BY NAME** vs
  `benchmarks/aarch64_md5_fixes_2026-08-07_final.tsv.zst` **CLEAN on all four**,
  and t=1-vs-t=8 CLEAN in both arms. The DEFAULT arm is not a formality —
  `ShardRecs::find` was edited.
* **tracker unit tests 42 pass** across default / `--no-default-features` /
  `__rect_1shard` / `__probe_wide` / `__probe_sites` / `zerocopy` /
  `__probe_lock_park` / `__bps_blocks` / `__msb_5`, one configuration at a time.
* **Teeth, tracker: 7 of 9 planted mutations FAIL the suite.** The two that do
  not are belt-and-braces duplicates of one cap, and removing BOTH fails. One
  mutation (rectangle-vs-rectangle comparing only row 0) passed the first time it
  was planted; the test grid was rewritten to signed offsets to catch it rather
  than the finding softened.
* **Teeth, decoder:** planting `rect.row(row + 1)` gives **277 mismatches of
  358**; restored gives 358/358 clean.
* **Instrument proven to see the code before any clock:** registrations
  569,690 → 409,349 per frame (`probe-sites`, `lost = 0`), `n_rect` 22,605/frame,
  `n_rect_declined` **0**, and `w_shards = w_blocks = w_full = 0` — identical to
  base on every cell, so the rectangle never promotes to the wide path.
* **`forbid(unsafe_code)` proven ACTIVE**, not read: a planted
  `unsafe { transmute }` in `src/picture.rs` fails the build at `lib.rs:13:12`;
  restored, sha256 identical. No `unsafe` added to `rav1d-safe`. The two new
  `unsafe` uses are in `rav1d-disjoint-mut` and each carries a safety comment
  naming the invariant.
* **Miri, both aliasing models × {default, `__rect_1shard`}**, one target at a
  time. Stacked Borrows / default features — the shipping configuration — is
  clean on every non-timeout target, `--lib` **42 passed** including all 13 new
  rectangle tests. `shard_liveness` TIMES OUT locally as
  `docs/AGENT_BRIEF.md` warns and is reported as a timeout, never as green; CI's
  Linux Miri legs cover it. Full matrix in `docs/RECT_RECORDS.md` §6c.
* `cargo fmt --all --check` rc=0; 6 clippy `-D warnings` legs rc=0.
* **Pre-existing on 140f914, verified by running the same commands there:**
  `__shards_1` fails the same 5 tests BY NAME, `--all-features` does not compile,
  its clippy leg has 8 errors, and root `clippy --release --all-targets` fails on
  `_dev` examples.

## What is NOT done

* The five CDEF / `cdef_apply` sites the counterfactual names (7.27-8.00 rows on
  **1.000** shards each) were **not wired**. `docs/RECT_RECORDS.md` §7b prices
  them at ~2.7% of the tracker and names the ten-minute arm
  (`RAV1D_CDEF_DOUBLE`) that settles it without implementing anything.
* No mutable rectangle ships (`add_rect_mut` is `#[cfg(test)]`).
* One box (M4 Pro, aarch64), 8-bit 4:2:0 in the timed grids, no x86_64 — where
  the multi-shard rectangle's extra lock traffic would be dearer, not cheaper.
* The t=1 code-size cost is attributed by elimination, not measured: no
  `cargo asm` or `llvm-lines` was taken.

**Do not merge** — measurement round, for review.


lilith:
### CI green, including both Linux Miri legs

Commit `2ef7742` (code-identical to HEAD — everything after it touches only
`docs/*.md` and a benchmark TSV) is green on both workflows:

* **`ci.yml`: 35 distinct checks, all pass** — same count as #503/#504, so the
  full matrix fired rather than the path-filtered subset. Includes
  `windows-11-arm`, `macos-15-intel`, `ubuntu-latest i686`, both conformance
  permutation legs, Tile Threading, and Threading race gates.
* **`disjoint-mut-ci.yml`: 13/13 jobs pass**, including
  **Miri (Stacked Borrows) 54m46s** and **Miri (Tree Borrows) 1h25m54s** over the
  whole package with `--no-fail-fast` — the legs that cover the
  `shard_liveness` target which times out locally on aarch64, and the ones that
  matter most here since the reverted March-2026 strided tracker's UB was
  Miri-only.

The `Documentation` failure on the first push was mine — two intra-doc links from
public items into the private `checked` alias — fixed in `2ef7742` and reproduced
locally with the exact CI command first.

Local Miri (aarch64, both models × {default, `__rect_1shard`}, one target at a
time) is clean on all 7 non-timeout targets in all four columns;
`shard_liveness` times out in all four and is reported as a timeout, never as
green. Full matrix: `docs/RECT_RECORDS.md` §6c,
`benchmarks/rect_records_miri_2026-08-11.tsv`.

PR #506, opened 2026-08-11
The rectangle's +1% at t=1 is CODE PLACEMENT, not the mechanism — and the CDEF sites priced

lilith:
## The `+1%` at t=1 that kept the rectangle default-off is CODE PLACEMENT, not the mechanism

**Nothing ships.** The default is unchanged and *verified* byte-equivalent to
`main`'s build (`__text` 1,839,536 → 1,839,536, 0 symbols resized, 0 added).
This PR is the attribution, the instruments that produced it, and the CDEF
pricing arm #505 §7b asked for.

### The finding

#505 measured **+1.0% to +1.3% wall at t=1 on `v4k8tile`, 0 of 11 rounds, in two
sessions**, in an arm where the rectangle path provably never executes, and
attributed it to code size **by elimination** — no `cargo asm`, no
`cargo llvm-lines`, and a layout control (`plainC`, same source, second
worktree) that moves no symbol and therefore read ±0.1%.

Static instruments first, no clock:

* Every hot 8bpc loop-filter symbol has a **byte-identical instruction stream**
  between the arms (`scripts/perf/text_symbol_diff.sh`). The only executed
  function whose codegen changed is `LfBlock::open`: **+8 bytes, +2
  instructions** at 8bpc.
* `cargo llvm-lines`: **+0.51% IR**, 81% of it `fill_rect`'s twelve
  monomorphisations — cold at t=1.

Then the control the campaign did not have (`loopfilter::text_pad`,
`src/text_pad.rs`): N KiB of `#[used]` functions nothing can call, in a build
where `text_layout_diff.py` reports **0 symbols resized** and a planted `panic!`
in the pad **leaves the md5 unchanged**, so it provably never runs.

`v4k8tile` t=1, n=11, idle, ratios vs `main`'s binary:

| arm | added `__text` | codegen changed? | wall | sign |
|---|---|---|---|---|
| byte-identical copy | 0 | — | **1.0006** | **4/11** |
| **dead text** | +4,828 | **no symbol resized** | **1.0110** | **0/11** |
| dead text | +9,692 | no symbol resized | 1.0117 | 0/11 |
| dead text | +14,556 | no symbol resized | 1.0109 | 0/11 |
| dead text | +19,420 | no symbol resized | 1.0156 | 0/11 |
| #505's rectangle arm | +13,700 | `open` +8 B | 1.0147 | 0/11 |

Nine binaries differing from `main`'s by **+1,132 B to +19,420 B** — dead code,
a pure refactor that *shrinks* `LfBlock::open` by 17%, near and far modules —
all land in **+1.1% to +1.6%** and are mutually within ±0.4%.

**Against a same-source control the rectangle costs nothing at t=1: 0.9967
(7/9).** The tax scales with working set: **+1.4% at 4K, +0.7% at 1024×576,
zero at 256×2048**.

So the acceptance bar for this round — t=1 inside ±0.5% of `main` with a
coin-flip sign at n≥11 — **is not met by any arm, including arms that change no
executed instruction anywhere**. Per the brief, the default was not flipped.

### The t=8 win, re-measured at the shipped configuration

Never inherited from the gated arm. `ship` = the restructure + `__lf_rect`;
`plain2` is the same source without it, which isolates the rectangle from the
layout draw.

| cell t=8 | wall vs `plain2` | sign | CPU | sign | vs `main` |
|---|---|---|---|---|---|
| `c1024x192` | **0.9851** | **12/13** | 0.9922 | 9/13 | 0.9900 |
| `c1024x384` | **0.9762** | **12/12** | **0.9845** | **12/12** | 0.9924 |
| `c1024x576` | **0.9826** | **11/11** | **0.9844** | **11/11** | 0.9925 |
| `text_q20` | 0.9962 | 8/13 | **0.9739** | **13/13** | 0.9923 |
| `ui_q20` | 0.9930 | 7/12 | 0.9929 | 9/12 | 0.9988 |
| `c3840x256` | 0.9981 | 7/13 | 0.9964 | 9/13 | 1.0063 |
| `c256x2048` | 1.0015 | 5/11 | 0.9962 | 8/11 | 1.0075 |
| `v4k8tile` | 0.9973 | 8/12 | 0.9951 | 11/12 | 1.0058 |

Control (byte-identical copy of `plain2`) 0.9962–1.0060, signs 3/13–8/13, every
cell. **Narrower than #505's B3 grid** — `c3840x256` was 0.9867 (6/7) there and
is null here — with stronger signs where it wins. The `vs main` column is the
shipping problem in one column: the layout draw eats about half the win and
flips three cells adverse.

### Task 2 — the CDEF sites priced without implementing anything

`RAV1D_CDEF_DOUBLE` (`--features __probe_cdef_double`) doubles the five CDEF
registration sites in ONE binary, immutably at the read sites and mutably at the
write site, dropped before the real guard — the `RAV1D_LF_DOUBLE` shape.

| cell t=8 | +regs/frame | wall | sign | CPU | sign | **ns/registration** |
|---|---|---|---|---|---|---|
| `c1024x576` | +121,856 (23.0%) | **1.0409** | **0/12** | **1.0386** | **0/12** | **5.27** |
| `c256x2048` | +159,424 (28.0%) | 1.0134 | 1/13 | **1.0223** | **0/13** | **3.27** |
| `text_q20` | **+0** | 1.0000 | 5/12 | 1.0009 | 5/12 | — (null control) |

Against `LfBlock::fill`'s 2.42–2.71 ns and `c256x2048`'s 19.71 ns/registration
*average*. On `c256x2048` the sites are **4.6% of the tracker** (§7b guessed
2.7%) — small. **On the 1024-wide family they are +4.09% of WALL**, four times
§7b's "under ~1%, nothing to win", i.e. an arithmetic ceiling of ≈3.6% for an
8-rows-to-1 collapse, on exactly the cells where the `fill` rectangle already
pays −1.7% to −2.4%. Standing caveat: `fill`'s own doubling priced +3.37% wall
on `c256x2048` and its actual collapse delivered ~0 there.

### Consequence for the campaign

**"The same source built in a second worktree" is not a layout control.** It
moves nothing and reads ±0.1%. Any t=1 claim about added code on this box must
be judged against a *same-source* arm, or against `--features __pad_text` /
`__pad_far`, and `main`'s current 4K t=1 number is ~1.1% better than a typical
layout for this binary.

### Gates

* corpus **766 PASS + 2 SKIP** in **both** arms × t=1 and t=8, no
  `--skip-group`; set-diff **BY NAME** vs the committed baseline **CLEAN on all
  four**; t=1-vs-t=8 CLEAN in both arms.
* **default codegen byte-equivalent to the base commit** (the gate this branch
  most needs).
* 5 planted mutations, all caught — including the **pad-never-executes** proof
  (a `panic!` in the pad leaves the md5 unchanged) and a CDEF-count mutation
  that lands on its **exact** expected value (−36,864, that site's population).
* `forbid(unsafe_code)` proven **ACTIVE** at `lib.rs:13:12` by planting a
  transmute. No `unsafe` added; `crates/rav1d-disjoint-mut` untouched.
* clippy 8 legs `-D warnings` rc=0; `cargo fmt --all --check` rc=0.
* Miri (both models) is running and its table will be added to
  `docs/RECT_SHIP.md` §8c before this is ready to merge.

Full write-up: `docs/RECT_SHIP.md`. Record:
`benchmarks/rect_ship_*_2026-08-11.{tsv,txt,meta}`.


lilith:
### Miri and CI complete

**Miri, both models × both feature sets** (`benchmarks/rect_ship_miri_2026-08-11.tsv`, `docs/RECT_SHIP.md` §8c): 7 of 8 targets clean in all four columns (`--lib` 42/40, `soundness` 25, `narrow_release` 1, `wide_exclusion` 1, `guard_move_release` 2; `pic_buf_overflow` and `aligned_miri` select **0 tests** and are reported as 0). `shard_liveness` **TIMEOUT(900s)** in all four locally, reported as a timeout — CI's Linux legs, which do cover it, are green under **both** models (SB 53m37s, TB 1h21m42s).

**CI: 35 of 35 checks pass**, the #504/#505 count. Getting there needed one extra thing worth flagging: `disjoint-mut-ci.yml` is path-filtered to `crates/rav1d-disjoint-mut/**`, and this branch adds registration sites in `src/cdef_apply.rs` / `src/safe_simd/cdef_arm.rs` **without touching the tracker crate** — so 13 legs, including both Miri ones, could never fire. `e2b5fd7` adds `workflow_dispatch` to that workflow (and, because the file is in its own paths filter, that commit also made the legs run here). That is exactly the case the filter excludes and the case that most wants Miri.

**Not merging** — left open for review. The decision this PR puts in front of you is not "is the rectangle good" (it is free at t=1 against a same-source control and −1.5% to −2.4% wall at t=8 on the 1024-wide family) but "is a ~1.1% layout redraw on 4K t=1 worth it", and that tax is paid by *any* change to this binary, including a refactor that only shrinks code.

PR #509, opened 2026-08-12
perf(cdef 16bpc): fixed-length padding copies instead of memmove — split out of #490, NOT a measured win

lilith:
The one code hunk from PR #490, lifted verbatim from `be289f4` and rebased onto `3ed9bcd`
(`src/safe_simd/cdef_arm.rs` is unchanged on `main` since #490's base `5606efe`, so this is a clean
lift). The records half of #490 lands separately in **#508**; this PR exists so the code gets its
own decision, which is what #490 asked for: *"Do not merge without a maintainer read of §5's revert
rationale."*

## Mechanism — verified, independently of the wall clock

`padding_8bpc` copies rows through `widen_row` → `widen_n::<N>`, a compile-time trip count.
`padding_16bpc` needs no widening, so it was spelled `copy_from_slice` — a **runtime** length, which
lowers to a `_platform_memmove` call for 4 to 12 `u16`s, up to `h + 4` times per CDEF block.

* Before: **381 of 549** memmove samples at 1024x576 came from `cdef_filter_block_16bpc_inner`;
  8bpc CDEF contributed **zero**.
* After: memmove inclusive **636 → 201 samples (−66%)**, and `cdef_filter_block_16bpc_inner`
  disappears from the caller list.

## It is NOT a measured win — this is the case against

| cell | head/base | sign | p |
|---|---|---|---|
| `L64x36_420_10b` | **0.9782** | 9/9 | 0.004 |
| `L256x144_420_10b` | 1.0078 | 1/9 | 0.039 *(nominally a significant **regression**)* |
| **8bpc control** `L1024x576_420_8b` — code the change cannot execute | **1.0094** | 3/9 | 0.508 |

So the A/B's own floor is ~1%, six of eight 10bpc cells move less than that, and no cell is
band-disjoint. #506 has since generalised that floor to a **±1.5% code-placement lottery** on any
t=1 arm, which strengthens the case for treating everything but the 64x36 cell as null.

## Correctness

Byte-identical by construction — same bytes, same order, same destination, same borrow guards.
#490's gates: corpus **766/766 at t=1** and **753/753 at t=8**, set-diffed by name with the md5 as
the value; **teeth planted** (`copy_n` writing `a[0]` into its last lane) gave **71 MISMATCH of 71**
on `10-bit/data` while `8-bit/data` stayed **358 PASS**, proving the path live *and* 16bpc-only,
then restored byte-exact.

## Recommendation

Ship for consistency with `padding_8bpc`, or close — both are defensible on these numbers, and
`docs/CONTENT_CLASS.md` §5 (landing in #508) records the measurement either way. I have not
re-measured: another agent holds the box for timed runs.

### DIFF
--- a/src/safe_simd/cdef_arm.rs
+++ b/src/safe_simd/cdef_arm.rs
@@ -162,6 +162,42 @@ fn widen_row(dst: &mut [u16], src: &[u8], n: usize) {
     }
 }
 
+/// `dst[..N] = src[..N]`, with a compile-time trip count.
+///
+/// The 16bpc twin of [`widen_n`]. There is no widening to do, so the obvious
+/// spelling is `copy_from_slice` — but that takes a RUNTIME length and lowers
+/// to a `memmove` libc call, which for 4 to 12 `u16`s is all call and no copy.
+/// With `[u16; N]` the move is a pair of `ldr q`/`str q` inline.
+#[inline(always)]
+fn copy_n<const N: usize>(dst: &mut [u16], src: &[u16]) {
+    let a = <&[u16; N]>::try_from(&src[..N]).unwrap();
+    let d = <&mut [u16; N]>::try_from(&mut dst[..N]).unwrap();
+    *d = *a;
+}
+
+/// Dispatch [`copy_n`] over the same length set [`widen_row`] handles.
+///
+/// Why this exists: `padding_8bpc` went through `widen_row` and so had no libc
+/// traffic at all, while `padding_16bpc` spelled the same copies as
+/// `copy_from_slice`. `docs/SIZE_SWEEP.md` Q2 profiled the consequence —
+/// **381 of 549 `_platform_memmove` samples at 1024x576 came from
+/// `cdef_filter_block_16bpc_inner`, and 8bpc CDEF contributed zero** — and
+/// named it as a distinct item from any missing kernel. Up to `h + 4` of these
+/// calls happen per CDEF block (one per block row, two top, two bottom).
+///
+/// Byte-identical by construction: same bytes, same order, same destination.
+#[inline(always)]
+fn copy_row_u16(dst: &mut [u16], src: &[u16], n: usize) {
+    match n {
+        4 => copy_n::<4>(dst, src),
+        6 => copy_n::<6>(dst, src),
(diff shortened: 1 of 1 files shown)


lilith:
Merging — on the **mechanism**, explicitly NOT on the wall-clock, and the PR is right that the wall-clock is not a win.

**Why the timing grid cannot decide this.** The 8bpc control **cannot execute the changed code** and still reads **+0.94%**. #506 since established why: a ±1.5% code-placement lottery underlies t=1 measurements — provably-dead, never-executed padding costs the same +1.1% to +1.6% as a real feature. So `L64x36_420_10b`'s 0.9782 (9/9, p=0.004) and `L256x144_420_10b`'s +0.78% (1/9, p=0.039) both sit inside noise this grid demonstrates it has. The p-values are computed against a within-arm spread that does not include the between-binary layout draw, which is the term that dominates here. Not a criticism of the measurement — it is why the PR correctly declined to claim a win.

**What IS measured, and is not layout-dominated:** `docs/SIZE_SWEEP.md` Q2 profiled **381 of 549 `_platform_memmove` samples at 1024x576 inside `cdef_filter_block_16bpc_inner`, with 8bpc CDEF contributing zero**. That is a direct attribution of libc traffic to this function, not a wall-clock difference between two binaries, so the lottery does not touch it. Up to `h + 4` such calls per CDEF block.

**Three further reasons that do not depend on any timing:**
1. **Byte-identical by construction** — same bytes, same order, same destination. Zero output risk.
2. **Consistency**: `padding_8bpc` already routes through `widen_row` and has no libc traffic; `padding_16bpc` spelled the same copies as `copy_from_slice`. This makes the twins agree, which is a maintenance property independent of speed.
3. It is the **documented project pattern** — the fixed-size-array rule in the performance guidance exists precisely because a runtime length lowers to a call while `[T; N]` inlines, and `widen_n`'s own doc in this file already cites it.

Removing a profiled libc call from a hot path, byte-identically, to match the sibling path and the house pattern, is worth doing at a wall-clock delta this grid cannot resolve in either direction. If a later run on a layout-stabilised build shows a real regression, revert it then — the records landed in #508 are the evidence for that conversation.

PR #511, opened 2026-08-12
test(bounds): the standing guard-extent gate — split out of #489, CI must re-validate its ceilings

lilith:
**Stacked on #510** (the records half of #489) so the doc edits compose. Do not merge before it.

The code half of PR #489: `include/dav1d/picture.rs::note_pic_extent` + `PIC_EXTENT_CEILINGS`,
`tests/guard_extent_budget.rs`, and the CI `extent-gate` job. It deletes the **NOT LANDED** banner
#510 puts on `docs/BOUNDS_MAP.md` Part 2 → "The standing assertions", and turns on the ENFORCED / §0
sentences in `OWNERSHIP_MODELS.md` and `AGENT_BRIEF.md`.

## Why it is separate from the records

Unlike the rest of Part 2 this is **live code**. Dead in a default release build — no counter, no
atomic load, no branch, it is `#[cfg(any(debug_assertions, feature = "probe-sites"))]` — but
**active in every `debug_assertions` build and in CI**. And its `PIC_EXTENT_CEILINGS` constants were
measured on `675a7ff` over **575,577,925** reservations across 1,734 frames. `main` is twenty commits
past that, including **#497** (narrowed the x86 loop-filter V window), **#500/#501/#503** (shard
granularity) and **#505** (the strided rectangle).

**I have not re-run the corpus** — another agent holds the box for timed work. **CI is the
re-validation.** A red debug or coverage leg here means a ceiling needs re-measuring, not that the
gate is wrong; that is exactly the signal this PR exists to produce, and it is why it is not riding
the records landing.

## What it asserts

At the one funnel every tracked picture-plane reservation passes through
(`Rav1dPictureDataComponent::{slice_as, mut_slice_as}`):

1. a partial reservation may not span more than **one picture row** while tile threading is active —
   #469's and #475's defect;
2. it may not exceed its file's `PIC_EXTENT_CEILINGS` entry, or one picture row where the file has
   no entry — #485's, and the V-batch's.

**Extent, and not `reserved <= footprint`**, on purpose: at t=8 the ratio is already 1.000 at every
hot site, and two of the three refuted attempts kept `reserved == footprint` *while widening*. A
ratio assertion would have passed both.

## Teeth (from #489's mutation table, not re-run here)

| plant | result |
|---|---|
| `RAV1D_LF_HULL=1` (#475's shape, switch already in-tree, no source edit) | **17 in-decoder panics**, `spans 16 rows` |
| LF per-row read `W → W + 96` (the V-batch's shape, inside one row) | **FAILS** on the file ceiling (32 B), in release **and** in a plain debug `cargo test --test mt_stress` |
| `unsafe` block in `tile_threading_active()` | **fails to compile**, anchored on `lib.rs:13` |

The mutation also found a hole in the gate's own test flag — `bytes > ceiling && rows > 1` missed a
one-row widening, the V-batch's exact shape — fixed in #489's `7a5a910`.

## Does not collide with #505

`__lf_rect` registers through `dm().index_rect_as(..)`, not through the `slice_as` funnel this
hooks, so a strided rectangle record is not seen by rule 1. The only conflict against `main` was one
region of `include/dav1d/picture.rs` where #505/#506 added `rect_hull_arm` / `cdef_double_reads`
exactly where the ceilings table goes; both sides kept.


lilith:
**Re-validation result: the gate is GREEN on current `main`, and non-vacuous.** The `Guard-extent gate (bounds map invariant)` leg ran for 3m01s and printed its own counters:

```
extent budget: checks=588521423 checks_tile_threaded=588521423 max=3840 B
               max_under_tile_threading=3840 B whole_component_exempt=6313610
               fallback_ceiling=64 B frames=1734 undecodable_vectors=6
```

**588,521,423 reservations checked, every one of them under tile threading, over 1,734 frames** — against #489's 575,577,925 on `675a7ff`, so the population grew with `main` and the ceilings still hold. That is the corpus re-measurement this PR said it needed and that I could not run locally (another agent holds the box for timed work). The three teeth from #489's mutation table were not re-run; the liveness half is what these counters demonstrate.

All 23 checks pass. Merging.

PR #517, opened 2026-08-13
test(cdef): land #448's 12bpc 4:2:2 regression vector — the fix is on main, ungated

lilith:
## What this is

**#448's one-line fix is already on `main`. Its regression vector is not.** So a real
wrong-pixels defect — aarch64 16bpc CDEF selecting `pri_tap` from bit 0 of `pri_strength`
instead of bit `bitdepth_min_8` — is fixed and **ungated**, and has been for a week.

This lands the vector (245 B) plus the per-frame oracle test. Nothing else.

## The fix survived two refactors, verified by reading `main`

`main` now derives `pri_tap` in two places in `src/safe_simd/cdef_arm.rs`, and **both** shift:

* `filter_params():639` — `pri_tap: (4 - (pri_strength >> bitdepth_min_8 & 1)) as i16`.
  This is the one that **executes**: `cdef_filter_block_16bpc_inner` builds `params` and
  hands them to `cdef_filter_block_16bpc_neon` via `..._dispatch`.
* `cdef_filter_block_scalar():968` — the same expression. Its own doc comment says it is
  "reachable only if `Arm64::summon()` fails, which it cannot on a NEON-mandatory aarch64
  target".

#448's diff touched `cdef_filter_block_16bpc_inner`, which the refactor split into those two.

## Teeth — planted, observed, restored

Per `AGENT_BRIEF.md` §4, a green test proves nothing until a mutation makes it red.
`cargo test --release --test decode_md5_committed`, aarch64 macOS:

**Mutation at the executed site (`filter_params:639`), reverting to `4 - (pri_strength & 1)`:**

```
thread 'cdef_12bpc_422_matches_c_oracle_per_frame' panicked at tests/decode_md5_committed.rs:333:5:
12bpc 4:2:2 decode diverged from the aomdec/dav1d oracle on aarch64 (issue #446):
frame 1: expected 9f2291bf3f75dd440bc1d64ae26e0ac8, got 5c69666dc939fc1dc2a458cb3b61e917
frame 2: expected d4c7026d6bf22d5b973d59adbe73043d, got 95a1a33d2acce7339b05aa40e54e3321
frame 3: expected 484033dc917f8c49aba051070733e3ce, got fae265054b46b89dead0387e584863ab
frame 4: expected af113a069ef6a383e80522ce50943a99, got 6fbc7b83c599efb9909110329191736b
```

4 of 5 shown frames, which is exactly what #448's body reported ("4 of 5 frames in the
reproducer diverged, not 1"). Restored → **4 passed; 0 failed** for the whole file, 0.02 s.

**Mutation at `cdef_filter_block_scalar:968` is correctly INERT** — the test still passes.
That is the documented-unreachable copy, not a hole in the gate. Recorded here so the next
reader who plants a mutation at the first `pri_tap` they `grep` does not conclude the test
is toothless.

## Liveness, added on top of #448's version

8-bit is immune to this defect by construction (`bitdepth_min_8 == 0`, so bit 0 *is* the
parity bit). A future re-minimisation or container change that quietly produced an 8-bit or
4:2:0 stream would leave the MD5s checked and issue #446 ungated — a vacuous gate, the
family `AGENT_BRIEF.md` §4 warns about. The test now asserts every shown frame is
**12bpc / I422 / 64x64** before hashing.

## Deliberately NOT landed

#448 also carries **`examples/obu_yuv_dump.rs` (277 lines)**. That is code, not a record, and
a records triage is the wrong place for it. It stays on `fix/446-arm-cdef-highbd-pri-tap`
@ `6da3a29`. The test here is self-contained and does not need it.

## Scope

Test-only: one 245 B vector + 71 lines in an existing test file + a 4-line import/liveness
change. No decoder code touched. `cargo fmt -p rav1d-safe --check` clean.

Refs #446. Supersedes the gate half of #448.


lilith:
## Tier verification: the executed CDEF tier is the aarch64 NEON one, proved by a planted `panic!`

Follow-up to the mutation evidence in the body, because "the mutation at :968 was inert" needs a
non-inferential explanation. The scalar copy in `cdef_arm.rs` is not merely *probably* unreached —
it is provably never entered on this vector:

Planted at the top of `cdef_filter_block_scalar<BD>` (`src/safe_simd/cdef_arm.rs:965`, i.e. before
its own correct `pri_tap` line):

```rust
fn cdef_filter_block_scalar<BD: BitDepth>(
    ... ,
) {
    panic!("PLANTED: scalar CDEF fallback executed");
    let stride = dst.pixel_stride::<BD>();
```

Result — the test still passes, so that function is never called:

```
running 1 test
test cdef_12bpc_422_matches_c_oracle_per_frame ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.00s
```

Restored; `git diff HEAD --exit-code` clean, whole file green again 4/4.

That closes the tier question from both ends:

| site | role | mutation planted | test |
|---|---|---|---|
| `filter_params:639` | builds `FilterParams` for `cdef_filter_block_16bpc_neon` | `>> bitdepth_min_8` removed | **FAILS**, 4 of 5 frames |
| `cdef_filter_block_scalar:968` | documented-unreachable fallback | `>> bitdepth_min_8` removed | passes (inert) |
| `cdef_filter_block_scalar:965` | same function | `panic!` at entry | passes → **never called** |

So the tier under test is the NEON one that carried the bug, not a scalar path wearing its name.
`filter_params` is shared with the 8bpc NEON path, where `bitdepth_min_8 == 0` makes the two
expressions identical — the mutation is a no-op there by construction, which is exactly why 8-bit
was immune to the original defect.

Vector: **245 bytes**, `sha256 c99016a00b5ab4bc5da1ba160212ad0580b3c0c5aeb86848a9c8bd3aa4396abf`
— two orders of magnitude under the 30 KB commit ceiling.

Issue #433, opened 2026-07-12
fuzz: rav1d-safe::differential_dav1d crash (84524ec01770c762) — RECURRED after https://github.com/imazen/rav1d-safe/issues/415


lilith:
Found by the continuous fuzz farm (libFuzzer).

**This signature was previously filed and closed as fixed at https://github.com/imazen/rav1d-safe/issues/415 — crashes with the identical panic location kept arriving afterward, so this is either an incomplete fix or a re-break, not a duplicate.**

- **Crate**: `rav1d-safe`
- **Target**: `differential_dav1d`
- **First seen on arch**: `arm64`
- **Signature** (`84524ec01770c762`): `panic fuzz_targets/differential_dav1d.rs:192:13`
- **Repro + metadata in R2**: `s3://zenfuzz/crashes/rav1d-safe/differential_dav1d/arm64/84524ec01770c762/`
- **Local mirror**: `/mnt/v/fuzzes/_farm/crashes/rav1d-safe/differential_dav1d/arm64/2271c8fb62186ea5`

Reproduce:
```
cd rav1d-safe
# copy the repro from the R2/local path above, then:
cargo +nightly fuzz run differential_dav1d <repro-file>
```

<details><summary>Captured trace (tail)</summary>

```
/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0x39e438) (BuildId: 2823c5a845b10d9e31ef8a0e9c4d99ccb9976298)
    #15 0xaf1e608d9760  (/mnt/fuzz/work/zen/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0x469760) (BuildId: 2823c5a845b10d9e31ef8a0e9c4d99ccb9976298)
    #16 0xaf1e608df61c  (/mnt/fuzz/work/zen/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0x46f61c) (BuildId: 2823c5a845b10d9e31ef8a0e9c4d99ccb9976298)
    #17 0xaf1e60f009a0  (/mnt/fuzz/work/zen/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0xa909a0) (BuildId: 2823c5a845b10d9e31ef8a0e9c4d99ccb9976298)
[...]

Issue #426, opened 2026-07-07
fuzz: rav1d-safe::differential_dav1d crash (f5bf5f9d3e1c2dde) — RECURRED after https://github.com/imazen/rav1d-safe/issues/400


lilith:
Found by the continuous fuzz farm (libFuzzer).

**This signature was previously filed and closed as fixed at https://github.com/imazen/rav1d-safe/issues/400 — crashes with the identical panic location kept arriving afterward, so this is either an incomplete fix or a re-break, not a duplicate.**

- **Crate**: `rav1d-safe`
- **Target**: `differential_dav1d`
- **First seen on arch**: `x86_64`
- **Signature** (`f5bf5f9d3e1c2dde`): `panic fuzz_targets/differential_dav1d.rs:99:13`
- **Repro + metadata in R2**: `s3://zenfuzz/crashes/rav1d-safe/differential_dav1d/x86_64/f5bf5f9d3e1c2dde/`
- **Local mirror**: `/mnt/v/fuzzes/_farm/crashes/rav1d-safe/differential_dav1d/x86_64/5464037781fd2a2a`

Reproduce:
```
cd rav1d-safe
# copy the repro from the R2/local path above, then:
cargo +nightly fuzz run differential_dav1d <repro-file>
```

<details><summary>Captured trace (tail)</summary>

```
7736e619fa)
    #15 0x5f0aec8cf308  (/root/work/zen/rav1d-safe/fuzz/target/x86_64-unknown-linux-gnu/release/differential_dav1d+0x6ce308) (BuildId: da2b2c54e61466fa09cfbe1a478edc7736e619fa)
    #16 0x5f0aec8ccc25  (/root/work/zen/rav1d-safe/fuzz/target/x86_64-unknown-linux-gnu/release/differential_dav1d+0x6cbc25) (BuildId: da2b2c54e61466fa09cfbe1a478edc7736e619fa)
    #17 0x5f0aec8d1a8a  (/root/work/zen/rav1d-safe/fuzz/target/x86_64-unknown-linux-gnu/release/differential_dav1d+0x6d0a8a) (BuildId: da2b2c54e61466fa09cfbe1a478edc7736e619fa)
    #18 0x5f0aec8d5f40  (/root/work/zen/rav1d-safe/fuzz/target/x86_64-unknown-linux-gnu/release/differential_dav1d+0x6d4f40) (BuildId: da2b2c54e61466fa09cfbe1a478edc7736e619fa)
[...]

Issue #425, opened 2026-07-06
fuzz: rav1d-safe::differential_dav1d crash (644b67a11afd7c5f)


lilith:
Found by the continuous fuzz farm (libFuzzer).

- **Crate**: `rav1d-safe`
- **Target**: `differential_dav1d`
- **First seen on arch**: `arm64`
- **Signature** (`644b67a11afd7c5f`): `panic fuzz_targets/differential_dav1d.rs:130:17`
- **Repro + metadata in R2**: `s3://zenfuzz/crashes/rav1d-safe/differential_dav1d/arm64/644b67a11afd7c5f/`
- **Local mirror**: `/mnt/v/fuzzes/_farm/crashes/rav1d-safe/differential_dav1d/arm64/3302a46a4cd70ab5`

Reproduce:
```
cd rav1d-safe
# copy the repro from the R2/local path above, then:
cargo +nightly fuzz run differential_dav1d <repro-file>
```

<details><summary>Captured trace (tail)</summary>

```
/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0x46b8e8) (BuildId: 1b7ad3e20efcd69ff3f6451ef47e7fed2c1de8ac)
    #17 0xc3ce934b1fc4  (/mnt/fuzz/work/zen/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0x471fc4) (BuildId: 1b7ad3e20efcd69ff3f6451ef47e7fed2c1de8ac)
    #18 0xc3ce934b70cc  (/mnt/fuzz/work/zen/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0x4770cc) (BuildId: 1b7ad3e20efcd69ff3f6451ef47e7fed2c1de8ac)
    #19 0xc3ce93ad8d10  (/mnt/fuzz/work/zen/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0xa98d10) (BuildId: 1b7ad3e20efcd69ff3f6451ef47e7fed2c1de8ac)
    #20 0xc3ce93ad9a5c  (/mnt/fuzz/work/zen/rav1d-safe/fuzz/target/aarch64-unknown-linux-gnu/release/differential_dav1d+0xa99a5c) (BuildId: 1b7ad3e20efcd69ff3f6451ef47e7fed2c1de8ac)
[...]

PR #521, opened 2026-08-28
feat(managed): Settings::strictness — reject what libaom rejects, default Strict (#422, #424)

### LINKED ISSUE #422: Decodes desynced AV1 streams without error that aomdec rejects (Corrupted segment_ids) — masked a shipping encoder non-conformance
## Summary

rav1d-safe 0.5.7 decodes — without any error — AV1 still-frame streams that libaom's reference decoder rejects as corrupt. The streams come from a real encoder defect (zenrav1e#35, found via imazen/zenavif#29): the encoder omitted required chroma TUs for `BLOCK_16X4`/`BLOCK_4X16` blocks in 4:2:0, so the symbol stream desyncs at the first missing TU. `aomdec` fails the frame ("Corrupt frame detected / Corrupted segment_ids"); rav1d-safe returns a full decoded frame with no error and garbage pixels after the desync point.

This permissiveness has a real cost: zenavif's encode→decode round-trip pipeline uses rav1d-safe as its decoder, so **every 4:2:0 conformance break sailed through round-trip testing** — the encoder bug shipped through multiple sweep programs and was only caught when a sweep added an explicit aomdec gate (PALCONF). A decoder error (or at least a detectable signal) on this stream class would have caught it on day one.

## Repro

Artifact (4267-byte single-frame low-overhead OBU stream, base64): https://gist.github.com/lilith/e71c71d43b62316b99ac4a620bcce28b

```
base64 -d wfail.obu.b64 > wfail.obu    # md5 88cfb5cc9f4ce69a635bd06e1e7df23e
# reference decoder:
aomdec --summary -o /dev/null wfail.obu
#   Warning: Failed to decode frame 1: Corrupt frame detected
#   Warning: Additional information: Corrupted segment_ids
# rav1d-safe (zenavif's examples/ivf_raw, i.e. Decoder::decode on the raw OBUs):
ivf_raw wfail.obu out.raw              # exit 0, 73728 bytes (192x256 I420), no error
```

Producing more of the class (any number of variants): zenrav1e @ `b15bc35f` (last pre-fix master) or any master commit from `7d254289`..`b15bc35f`, encode any photo at 4:2:0 through the topdown partition path with `non_square_partition_max_threshold = BLOCK_64X64` + `partition_range (4,16)` (the ravif/cavif `-s2 --yuv 420` shape). 33/36 corpus cells in the discovery sweep decoded "cleanly" under rav1d-safe while 100% were aomdec-rejected.

## What the stream actually contains

For a `BLOCK_16X4` intra block at an odd mi row in 4:2:0, the encoder wrote **no chroma TUs at all** — conforming decoders parse a `TX_8X4` chroma TU there (`Subsampled_Size[BLOCK_16X4][1][1] = BLOCK_8X4`, spec 5.11.38). Everything after the first such block is a misaligned symbol stream. libaom happens to detect it via its segment_id range check (`segment_id > last_active_segid`); rav1d-safe evidently reads the same garbage without tripping any validation on this input.

## Data-dependent, not universal

rav1d-safe does NOT mask the whole class: a synthetic 256x256 stream from the same encoder bug (zenrav1e `tests/sliver_chroma_roundtrip.rs` pre-fix, `SLIVER_TEST_DUMP_IVF` hook) is rejected with `InvalidData`. So some desynced paths do hit validation — the corpus streams above just don't.

## Ask

Not asking to bit-match libaom's error strings — but a stream whose symbol decode wanders into inconsistent state (segment ids out of the signaled range would be one cheap detectable signal, matching dav1d/libaom behavior) should surface *some* error rather than silently returning garbage pixels. Happy to provide more samples from the sweep corpus if useful; the encoder side is already fixed (zenrav1e@17e67842) so this is purely about decoder strictness/validation parity.

Found 2026-07-03 during imazen/zenavif#29 triage.

### LINKED ISSUE #424: differential_dav1d: 10-bit decode diverges from dav1d on a mutated stream (not a SIMD-port bug)
## Summary

The `differential_dav1d` fuzzer (extended in 7892c420 to bit-exact-compare 10/12-bit planes) found a **deterministic 10-bit decode divergence** between rav1d-safe and dav1d 1.5.3:

```
DIVERGENCE Y row 129 col 22: rav1d=930 dav1d=892
```

Both decoders accept the stream and return a frame with **matching dims, pixel layout, and bit_depth (10)** — only the Y-plane pixels differ (930 vs 892, ~38 levels at 10-bit). Reproduces identically 3/3 runs.

## Triage

`__simd_test_log` dual-compute (SIMD vs scalar, threads=1) logged **zero SIMD-vs-scalar mismatches** during the decode. So this is **not** a SIMD-port divergence (unlike the #414 CDEF/ITX ARM class) — it's a decode-logic difference in a path both SIMD and scalar share, or a scalar-only path, that diverges from dav1d.

## Caveat (severity)

The input is a **fuzzer-mutated** stream — the 24 clean 10-bit seeds (cavif encodes, 420 + 444, q30/q75) did **not** diverge. So this may be an underspecified / non-conforming-stream corner rather than a clean-content decode bug. Confirming whether any *conforming* stream can reach the diverging path is the key next step before treating it as a decode bug.

## Reproducer

13278-byte raw AV1 OBU (didn't `tmin`-minimize — divergence needs the full frame). Block storage (needs libdav1d to reproduce, so not a git regression seed):

- `/mnt/v/fuzzes/rav1d-safe/differential/crash-b5bedd06f099c7cbdce4a16467b9b37f92b32443`
- sha256 `d204fcaa1c65bf7b915b98bba7e5d4ea4cd64da03056bb8c51c7ec9b98d2bf0a`

```
PKG_CONFIG_PATH=<libdav1d-1.5.3>/lib/pkgconfig \
LD_LIBRARY_PATH=<libdav1d-1.5.3>/lib \
cargo +nightly fuzz run --features differential differential_dav1d <repro>
```

## Meta

The 10/12-bit differential path was previously untested (`if rbd != 8 { return; }`) and `differential_dav1d` is not in the CI fuzz matrix (only `decode_obu`, `parse_seq_header`) since it needs libdav1d. Worth adding to CI (ubuntu-latest has libdav1d-dev >= 1.3) so this class is caught regularly.

lilith:
## What

`Settings::strictness: Strictness` — the decoder's conformance policy — with **`Strict` as the managed-API default**.

| variant | meaning |
|---|---|
| `Strictness::Lenient` | dav1d's library default (`strict_std_compliance = 0`): conceal and continue, bit-exact with dav1d even on corrupt input. |
| `Strictness::Strict` (default) | reject what the AV1 reference decoder rejects: every check dav1d gates behind `strict_std_compliance` (OBU trailing bits, tile symbol-decoder padding, forbidden bit, zero `timing_info`, `intra_only` + refresh `0xff`, `show_existing` of a non-showable frame, identity MC with subsampled chroma, frame-thread reference-error propagation) **plus** the AV1 §6.10.8 requirement that a decoded `segment_id` lies in `0..=LastActiveSegId` — dav1d resets it to 0 with a `// error?` comment; libaom errors with "Corrupted segment_ids". |

- `#[non_exhaustive]`, `Ord` (`Lenient < Strict`), `Default = Strict`.
- `strict_std_compliance: bool` is **deprecated, not removed** (API-snapshot rule); `true` still means `Strict`, the stricter of the two wins.
- C API unchanged: `Dav1dSettings.strict_std_compliance` 0 → `Lenient`, non-zero → `Strict`; the internal `Rav1dSettings` default stays dav1d's `Lenient`.
- Both real streams behind the issues are committed as `tests/strictness_vectors/` (4.3 KB + 13.2 KB) and gated by `tests/strictness.rs`: **Lenient decodes both, Strict → `Error::InvalidData`**. aomdec rejects both ("Corrupted segment_ids" / "Failed to decode tile data"); dav1d 1.5.3 `--strict 0` decodes both, `--strict 1` rejects both.

## Why `Strict` by default

rav1d-safe's job here is still images and verifying encoder output. zenrav1e#35 (missing chroma TUs in 4:2:0) desynchronised the symbol stream and **passed every encode→decode round trip** because the decoder concealed it — #422. #424 is the same class from the fuzzer: a mutated 10-bit stream that dav1d and rav1d-safe conceal *differently*, i.e. pixels the spec never defined. A still-image pipeline wants an error, not garbage.

## Cost, measured (`benchmarks/strictness_2026-08-28.meta`)

- **766-vector dav1d md5 gate: unchanged** (14/14) at the Strict default.
- 315 real AVIFs/OBUs (codec-corpus + zenavif/tests + bench): identical verdicts except `avif-conformance/invalid/corrupted_mdat.avif` — Strict rejects it, correctly.
- **53,484 AVIFs** (`/mnt/v/input/zensim` sweep encodes): **0 disagreements** (106,968 decodes).
- Per-block cost: one predictable compare on the segmentation path; everything else is per-OBU / per-tile.

## Fuzz plumbing

- `decode_obu` / `parse_seq_header` pin `Lenient` (maximal reach). **`differential_dav1d` now compares strict against strict** (dav1d `strict_std_compliance` on): the byte-exact compare covers streams both decoders accept, and an accept/reject asymmetry is a real finding. Every divergence the farm filed against the old lenient compare (#425, #426, #433) was a stream aomdec rejects — all 5,170 artifacts checked, `benchmarks/differential_triage_2026-08-28.meta`.
- `tests/fuzz_regression.rs` gained a fifth `default_lenient` entry point: production defaults now reject 7 mutated seeds before the kernel they guard, which would have dropped the reach floor 21 → 20; the new entry point keeps every guard live with the floor **unchanged**. `examples/crash_sweep.rs` mirrors it.
- `examples/strictness_sweep.rs` — the survey tool behind the numbers (`--jobs`, `--quiet`; AVIF / IVF / raw OBU).

## Things to spot

- **Naming**: `Strictness::{Lenient, Strict}`; alternatives considered `Conformance::{Conceal, Reject}`. A third level (e.g. "pedantic" checks that cost something) can be added later thanks to `#[non_exhaustive]`.
- **Default asymmetry**: managed API `Strict`, C API / `Rav1dSettings` `Lenient` (dav1d parity) — intentional.
- **Deprecate vs remove** the bool: 0.6.0 is unreleased and already breaking, so removal would be cleaner; kept because of the snapshot "no deleted public items" rule.
- `Settings::effective_strictness()` is `pub(crate)`.
- Pre-existing, not touched: `RUSTDOCFLAGS=-D warnings cargo doc` fails on four broken intra-doc links in `include/dav1d/picture.rs` (private-item links + `feedback_mt_safe_batching`); none of the new docs are involved.

## Gates

`tests/strictness.rs` 5 passed · `fuzz_regression` 3 passed · `decode_md5_verify` 14 passed (766 vectors) · `decode_md5_committed` 4 passed (+ `__simd_test` 3) · fmt clean · api-doc snapshots regenerated (first commit, no public item removed) · clippy: no finding in any touched file (this box's clippy 1.98 flags two pre-existing `looprestoration.rs` `chunks_exact` sites that CI's clippy does not, plus the pre-existing dead `token_test_lock` in the lib test target — the feat commit message's "clean" overclaims; corrected in the record).

## Commits on the branch

1. `docs(api)`: public-API snapshot regeneration (stale since the probe-*/shards-* feature families; no public item removed).
2. `feat(managed)`: the change.
3. `fuzz(differential)`: strict-vs-strict compare.
4. `test(decode_md5_committed)`: the committed MD5s pin `Lenient` — they are cross-arch parity references and one (`arm_itx_16x64_dc_rect2.obu`) is a fuzz-derived stream Strict rightly refuses; this is what the first CI run failed on, on every leg.
5. `docs(fuzz)`: the differential triage record + mirror/doc consistency.

Closes #422. Closes #424.

https://claude.ai/code/session_01NCDE1rkJXg3ugNWzmNaN4i


lilith:
## DisjointMut: what this PR carries, and the deductive review it triggered

**This PR contains no DisjointMut code.** Its only `rav1d-disjoint-mut` files were the regenerated `docs/public-api/rav1d-disjoint-mut.*` snapshots, which document 17 items that landed on `main` between 08-07 and 08-13 (`index_rect{,_mut}{,_as}`, `DisjointImmutRectGuard` / `DisjointMutRectGuard`, `declare_row_stride`, `set_parallelism` / `set_tile_concurrency`, the `probe_*` hooks). That regen now lives on `main` (274b4d8e) and `main` is merged into this branch, so the PR diff is only the strictness change.

**Those items were reviewed deductively against `main` @ 00ad667, and the proofs are in `crates/rav1d-disjoint-mut/AUDIT.md` ("2026-08-28 review")** — Lemma 1 block cover (any shared element puts both registrants in one shard), 2 in-shard serialisation (publish precedes unlock; `allocated` is a superset of the live bits), 3 exactness of `rect_decode` / `rect_hit_range` / `find_from_rect` (the hull is a prefilter, `refine` rescans exactly), 4 publish/release ordering (Release/Acquire pairing; the lock-free retire has one writer at a time), 5 the wide list (written under every active lock, re-read under a lock on every narrow path), 6 the rect guards (rows disjoint by `|stride| ≥ seg`, one `&mut` row at a time, `NonNull` base, bounds derived from the element length), and U — units.

**One real defect (Lemma U), fixed on `main` in 02660934:** `index_rect{,_mut}` registered the rectangle in **bytes** (`lo * size_of::<V>()`) while `index{,_mut}` register in **`T::Target` elements**. Identical on `u8` buffers — every decoder call site is `index_rect{,_mut}_as` over a `u8` plane, so the decoder was never exposed — but on `DisjointMut<Vec<u16>>` a rectangle over elements `{0,1,8,9}` was recorded as bytes `{0..4, 16..20}`, `index_mut(8..10)` over the same elements found no overlap, and two live `&mut` were handed out from safe code (plus false positives in the inter-row gap, plus a bound that refused the upper part of the buffer). `crates/rav1d-disjoint-mut/tests/rect_units.rs` is the counterexample: 6 of 8 fail on the old code (the two `should_panic` cases do not panic), 8 of 8 pass on the fix under `cargo test` and under Miri with Stacked and Tree Borrows. Whole crate suite green in all feature sets; the decoder gates are unchanged by construction.

The merge also brought `main`'s `as_chunks` conversion in `looprestoration.rs`, which is what the newer clippy on the runner had started flagging (the only red check on the previous head).

